PR #4, opened 2022-03-22
chore: set up Palette

anandaroop:
Paired with @jonallured on this.

Our goal is to get some Artsy basics wired up before Hackathon. This PR sets up [Palette](https://palette.artsy.net)

This was a lot more fiddly than we would have liked 😐 but basically:
- figured out through extensive trial and error what was the most recent combination of libs and types could be installed for **styled-components** & **styled-system**, basically this 👇🏽 

  ```sh
  yarn add \
  styled-components@4.0.3 @types/styled-components@4.0.3 \
  styled-system@5.1.5 @types/styled-system@5.1.15
  ```

  This is very close to what Force has installed, although the styled-component types and lib are out of sync over there, and I have chosen to synchronize them here 🤷🏽

  Any attempt to move to a more recent version of styled-components here has so far resulted in `tsc` type-checking  being broken, so I'm leaving this at styled-components v4.0.x for now

  (cc @damassi @dzucconi in case either of you know what's up with the versions/types problem noted above)

- set up the root-level theme provider plumbing, and ensured intellisense was working in at least VS Code 

  Screen Shot 2022-03-21 at 10 38 45 PM

- created a simple demo page for Palette components at `/palette`

  Screen Shot 2022-03-21 at 10 33 02 PM

- moved the previous Palette-for-Tailwind demo page to `/palette/tailwind`


damassi:
I wasn't able to immediately fix the versioning in my example app unfortunately. I imagine its fixable, but we'd probably have to go to palette and get everything straightened up there first. 

We should go ahead and setup SSR rendering for styled component before merge: https://github.com/damassi/relay-nextjs-example/blob/main/src/pages/_document.tsx#L30-L56

package.json
```diff
@@ -12,11 +12,14 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "@artsy/palette": "^19.1.3",
     "cookie": "^0.4.1",
     "date-fns": "^2.28.0",
     "next": "12.0.7",
     "react": "17.0.2",
     "react-dom": "17.0.2",
+    "styled-components": "^5.3.3",
```
damassi, line 21: We can keep the latest perf updates from styled-components; its just the types that need to be downgraded in order to fix intellisense.

PR #5, opened 2022-03-23
chore: hokusai setup with nodejs template and some custom config [WIP]

kajatiger:
` hokusai dev start` and `hokusai test` should already be working and the setup script as well ✅
`hokusai staging status` still does not work ❌
It gives the error: `ERROR: Rendering template raised error <class 'jinja2.exceptions.UndefinedError'> <message ''externalIngressAllowSourceIP' is undefined'>`


hokusai/production.yml
```diff
@@ -0,0 +1,176 @@
...
+          resources:
+            requests:
+              cpu: 200m
+              memory: 256Mi
+            limits:
+              memory: 1Gi
+          readinessProbe:
+            httpGet:
+              port: {{ project_name }}-http
+              path: /api/status
```
kajatiger, line 61: @dblandin and I added the api/status route to respond with 200 ok

dblandin:
> It gives the error: ERROR: Rendering template raised error <class 'jinja2.exceptions.UndefinedError'> <message ''externalIngressAllowSourceIP' is undefined'>

Ahh, I think you have to add [these lines](https://github.com/artsy/horizon/blob/main/hokusai/config.yml#L6-L7) to the hokusai/config.yml file:

```yaml
template-config-files:
  - s3://artsy-citadel/k8s/hokusai-vars.yml
```

joeyAghion:
Pushed a commit that tries to adapt our environment variable handling for the framework's defaults. Since Next.js has [a strong convention around environment variable loading order](https://nextjs.org/docs/basic-features/environment-variables#environment-variable-load-order) already, I propose that we put non-sensitive defaults in `.env.development` (which is committed), and copy sensitive shared values to `.env.local` (which is not). If developers want to specify further, non-shared overrides, they can do so in `.env.development.local` which is also not committed.

Bonus: added these simple setup commands to the README. (Eventually these can be handled by a setup script.)

joeyAghion:
I hadn't noticed that there already _was_ a `setup` script in place, so I've updated it to set up `.env.local` and `.env.development` as described above.

joeyAghion:
Latest 2 commits (with @ovasdi): 

* Added `yarn build` command to `Dockerfile`, which is necessary to compile assets when running in production mode.
* Switched from to internal routing at `forque.stg.artsy.systems`, since we agreed to access this via VPN for now.

We initialized the staging environment with these commands:

```
hokusai registry push --force
hokusai staging create
hokusai staging env set GRAVITY_URL="..." METAPHYSICS_URL="..." NEXT_PUBLIC_GRAVITY_URL="..." NEXT_PUBLIC_METAPHYSICS_URL="..." CLIENT_APPLICATION_ID=... CLIENT_APPLICATION_SECRET=... NODE_ENV=production
hokusai staging refresh
```

Then, we added a DNS entry [in Route53](https://us-east-1.console.aws.amazon.com/route53/v2/home#Home) for the new `forque.stg.artsy.systems` name.

joeyAghion:
Also note that we switched both `Dockerfile` and `.nvmrc` to use node v16, so that local development and deployed images operate consistently. I'm not married to any particular version, but the even numbered versions offer long-term support and we've had problems elsewhere when unrelated PRs unintentionally rebuild the application on an upgraded node.

PR #6, opened 2022-03-23
refactor(app): Make app consistent with other Artsy JS apps

damassi:
This PR does a couple things:
- moves source code into `src` folder to better scope things
- adds absolute path import support 
- colocates everything to match our other JS apps (page/app related components live with page/app folder; stories live with components alongside co-located tests, and so on)
- adds `.vscode/settings.json` so that VSCode can pick up configurations and enforce them across developer environments
- upgrades Next.js to latest (12.1)


.eslintrc.js
```diff
@@ -18,5 +18,6 @@ module.exports = {
       "error",
       { vars: "all", varsIgnorePattern: "^_" },
     ],
+    "testing-library/render-result-naming-convention": "off",
```
damassi, line 21: This rule is creeping into source-code

anandaroop (changes requested):
Thanks for jumping in!

> * moves source code into `src` folder to better scope things
> * adds absolute path import support
> * adds `.vscode/settings.json` so that VSCode can pick up configurations and enforce them across developer environments
> * upgrades Next.js to latest (12.1)

🙌🏽 Yess to all this

> * colocates everything to match our other JS apps (page/app related components live with page/app folder; stories live with components alongside co-located tests, and so on)

This one is problematic because it is an anti-pattern in Next.js terms.

In every Next app I've seen, the `/pages` directory is strictly lightweight [Page components](https://nextjs.org/docs/basic-features/pages) that make use of their built-in conventions for [filesystem-based routing](https://nextjs.org/docs/routing/introduction). (I'd say this is part of the delight of using Next).

Everything else tends to be imported from a `/components` directory.

Notably, this is true of [their docs](https://nextjs.org/docs/basic-features/layouts) as well as their [example apps](https://github.com/vercel/next.js/tree/canary/examples) — see https://github.com/vercel/next.js/tree/canary/examples/with-stripe-typescript for one of the more fleshed out examples.

Having anything other than routable pages under the `/pages` directory will cause problems such as:

- Inadvertently bundling and exposing all those files as URLs (which will generally be broken links)

  Even in this PR, if you fire up the dev server you'll find that you can "browse" to http://localhost:3000/artists/dedupe/components/list/__tests__/ArtistList.spec

- Increasing build times, since Next is interpreting each of those files nested under `/pages` as a reachable URL that might need to have data injected at build time for SSG, ISR etc

So we'll definitely want to walk this commit back, and some new conventions / developer education will certainly be in order.

damassi:
> This one is problematic because it is an anti-pattern in Next.js terms.

This is not a fixed thing in Next and Next.js accommodates all! It might not be a "convention" in terms of the examples, but that's because examples are by nature very small and don't operate at scale. 

For something like a mono-admin such as this, we'll want to keep things as close to colocated as possible for organizational reasons and for that, we can use next's `pageExtension` prop: 

https://nextjs.org/docs/api-reference/next.config.js/custom-page-extensions#including-non-page-files-in-the-pages-directory

anandaroop:
Mind blown! Good on Next, yet again.

My googling skills were inadequate to finding that page the docs. I tried so many search terms, lol.

That's good to know about. But we do still need to configure that right? (See my comment about tests currently being exposed)

damassi:
On it now 🙏

anandaroop:
Nice, that fixes the exposed URLs as well as (presumably) by build-time concerns 🎉

damassi:
Yeah, now rather than a few new conventions / requirements we'd need to spread around as well as some dangerous conditions we'd need to guard against in PR review -- namely, unintentional exposure -- it's just one: a viewable "page" must contain `.page.tsx` in the name, which is arguably clearer as it makes things quite unambiguous.

anandaroop:
Seeing a couple small regressions involving Jest and Storybook. Looks like quick fixes — I'll push up a commit.

anandaroop:
All set now, we'd have caught wind of that sooner if CI was hooked up. That's coming soon, on the #product-velocity todo list, I'm pretty sure.

anandaroop (approved):

PR #15, opened 2022-03-25
feat: Add cypress scripts and update ci

jonallured:
Lets see if this "just works" 🤞 

/cc @anandaroop @damassi @joeyAghion @kajatiger

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,7 @@
 version: 2.1
 
 orbs:
+  cypress: cypress-io/cypress@1
   hokusai: artsy/hokusai@0.7
   horizon: artsy/release@0.0.1
 
@@ -33,6 +34,13 @@ workflows:
 
       - hokusai/test:
           <<: *not_staging_or_release
+
+      - cypress/run:
+          yarn: true
+          start: 'yarn dev'
+          executor:
+            name: cypress/default
+            node: 14.19.0
 
       - hokusai/push:
           name: push-staging-image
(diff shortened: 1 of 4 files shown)


jonallured:
Ugh nothing is easy! So after much tinkering I got this back on track by using the cypress circleci orb HOWEVER the node versions aren't playing nicely:

```
error next@12.1.0: The engine "node" is incompatible with this module. Expected version ">=12.22.0". Got "12.19.0"
error Found incompatible module.
```

I think what this means is that the version of next we're using requires a newer version of node than the latest version of the cypress orb installs. Not too sure what to do about that. 🤔

damassi:
I believe you can set your own executor; maybe try pointing that at our node: https://circleci.com/developer/orbs/orb/cypress-io/cypress#usage-using-node14

Maybe something like:
- https://github.com/artsy/force/blob/main/.circleci/config.yml#L27 or 
- https://github.com/artsy/force/blob/main/.circleci/config.yml#L51

jonallured:
Ok progress!

https://app.circleci.com/pipelines/github/artsy/forque/106/workflows/7aae7435-037c-4658-bb3e-ef3ed41a1dbc/jobs/140

That's still red but now it's red for a sensible reason - the next.js server wasn't started. @damassi any suggestions on the best way to start the server?

jonallured:
Wow I figured it out and this PR is green! Going to assign you @damassi - lmk if you want anything in before this is merged or if I can do things in a follow up PR.

PR #16, opened 2022-03-25
feat: add some env debug pages

anandaroop:
While we figure out how to get the ENV vars to the expected places...


damassi:
Ooh, good idea 👍

PR #21, opened 2022-03-27
build: more experimentation with runtime env vars

joeyAghion:
Confirmed locally, running in a production-like way, that the runtime environment variables (from `publicRuntimeConfig`) are available if the initially loaded page was server-side rendered, but _not if the initially loaded page is client-side rendered_.

My steps:
```
yarn build
CLIENT_APPLICATION_ID=... CLIENT_APPLICATION_SECRET=... GRAVITY_URL=... METAPHYSICS_URL=... NEXT_PUBLIC_GRAVITY_URL=... NEXT_PUBLIC_METAPHYSICS_URL=... NODE_ENV=production yarn start
```
Then, visit `/env/server` and follow the `Link` to `/env` and back, and confirm that the public runtime variables are available. Start on `/env` (which is client-side), and follow the link to `/env/server` to confirm that they're _not_ available.

### DIFF
--- a/src/pages/env/index.page.tsx
+++ b/src/pages/env/index.page.tsx
@@ -1,6 +1,7 @@
 import type { NextPage } from "next"
 import Head from "next/head"
 import getConfig from "next/config"
+import Link from "next/link"
 
 const { publicRuntimeConfig } = getConfig()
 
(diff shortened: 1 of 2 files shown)


joeyAghion:
Another way of demonstrating this issue: Load `/env/server`, then follow the nav link to `/artists/dedupe`, and that page _works_. Coming from any other statically generated location, and it doesn't work.

anandaroop (approved):

damassi:
Y'all 😄 please try using the 

```
env: {
  foo: process.env.FOO
}
```
setting in `next.config.js` and see what happens 🙏

joeyAghion:
As discussed, `env` didn't work either.

Even if we force pages that depend on runtime config to render server-side with `getServerSideProps`, that doesn't seem to supply the correct config when navigating from a statically generated page.

It seems we'd have to force _every_ page to be rendered server-side or create an API that serves the necessary configuration accurately to static pages.

PR #22, opened 2022-03-28
fix(dedupe): force SSR, to make runtime config available

anandaroop:
Follow up to https://github.com/artsy/forque/pull/21 as well as [this Slack comment](https://artsy.slack.com/archives/C0376CFU527/p1648405432752219?thread_ts=1648217677.423169&cid=C0376CFU527)

This adds a `getServerSideProps` to the dedupe app in order to force into SSR mode, which is what is required to [access the runtime config in the Next-blessed way](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration).


joeyAghion (approved):
👍 from me.

Based on my experimentation, won't this still lack the necessary config if navigated to from a statically generated page?

PR #25, opened 2022-03-29
Update pattern for prettier

jonallured:
We did some reorganization and forgot to update the pattern for prettier so this PR just adds the `src` prefix.


anandaroop (approved):
Sweet, thanks

PR #27, opened 2022-03-29
feat(auth): add relay authentication; fix initial login flash

damassi:
This updates our relay layer to support authenticated queries and additionally fixes the logged in page flash.

### DIFF
--- a/src/pages/_app.page.tsx
+++ b/src/pages/_app.page.tsx
@@ -1,25 +1,31 @@
 import "../styles/globals.css"
 import "regenerator-runtime" // relay network requirement
-import type { AppProps } from "next/app"
+import App from "next/app"
+import type { AppContext, AppProps } from "next/app"
 import { Theme, injectGlobalStyles } from "@artsy/palette"
 import { Layout } from "components/Layout"
 import { ErrorBoundary } from "system/ErrorBoundary"
 import { useEnvironment } from "system/relay/setupEnvironment"
 import { RelayEnvironmentProvider } from "react-relay"
+import { getUserFromCookie } from "system/artsy-next-auth/auth/user"
+import { NextApiRequest } from "next"
 
 const { GlobalStyles } = injectGlobalStyles(`
   /* overrides and additions */
 `)
 
-function MyApp({ Component, pageProps }: AppProps) {
-  const environment = useEnvironment(pageProps.relayData)
+export default function MyApp({ Component, pageProps }: AppProps) {
+  const environment = useEnvironment({
+    initialRecords: pageProps.relayData,
+    user: pageProps.user,
+  })
 
   return (
     <Theme theme="v3">
       <RelayEnvironmentProvider environment={environment}>
         <GlobalStyles />
         <ErrorBoundary>
-          <Layout>
+          <Layout user={pageProps.user}>
             <Component {...pageProps} />
           </Layout>
         </ErrorBoundary>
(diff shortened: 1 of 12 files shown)


src/system/relay/middleware.ts
```diff
@@ -1,20 +1,36 @@
...
-    ttl: 60 * 1000,
-  }),
-]
+export const getRelayMiddleware = (
+  user?: UserSessionData | null
+): Middleware[] => {
+  const authenticatedHeaders: Headers = user
+    ? {
+        "X-USER-ID": user.id,
+        "X-ACCESS-TOKEN": user.accessToken,
```
jonallured, line 19: I always get confused about why we have both of these headers - does the token include the id? What would be mean to have a mis-match? 🤔

jonallured (approved):
LGTM!! 🎉

src/system/relay/middleware.ts
damassi: It is MP weirdness. Some queries look explicitly for that id header, not sure why. I suspect its just old stuff

src/pages/_app.page.tsx
```diff
@@ -28,4 +34,9 @@ function MyApp({ Component, pageProps }: AppProps) {
   )
 }
 
-export default MyApp
+MyApp.getInitialProps = async (appContext: AppContext) => {
+  const appProps = await App.getInitialProps(appContext)
+  const user = await getUserFromCookie(appContext.ctx.req as NextApiRequest)
+  appProps.pageProps.user = user
+  return { ...appProps }
+}
```
anandaroop, line 42: Oh, so being part of `_app.tsx` — does this now force every page now to be ssr? I.e. I'm wondering if we don't have to recommend [this kind of paperwork](https://github.com/artsy/forque/pull/24/files) any more?

src/system/artsy-next-auth/auth/cookies.ts
```diff
@@ -16,8 +16,8 @@ export function setTokenCookie(res: NextApiResponse, token: string) {
   res.setHeader("Set-Cookie", cookie)
 }
 
-export function getTokenCookie(req: NextApiRequest) {
-  const cookies = parseCookies(req)
+export async function getTokenCookie(req: NextApiRequest) {
+  const cookies = await parseCookies(req)
```
anandaroop, line 20: `parseCookies` isn't async tho, right? Why is this await necessary?

damassi: To fake out the event loop and let it resolve correctly 😅

src/pages/_app.page.tsx
damassi: > if we don't have to recommend this kind of paperwork

I think you're right! (I was just assuming that every page would naturally have a root relay base query, so would be using it by default per the pattern)

src/system/artsy-next-auth/auth/cookies.ts
anandaroop: Sooo... if I remove this async/await and everything still seems to work... what am I missing? Or how can I try to make it break?

(It's a small point, but just in the interest of keeping the hacks to the bare minimum necessary)

damassi: If you removed it I _think_ what you would see is [this](https://artsy.slack.com/archives/C0376CFU527/p1648572501052449?thread_ts=1648567911.436469&cid=C0376CFU527) but only after the SSR pass. 

Page loads fine, click to A, then click back to home, triggers error.

PR #28, opened 2022-03-29
feat(user): Stub out user page 

damassi:
This adds a new user page to Forque. Next PR will stub out all of the UI while @jonallured works on some of the MP bits for `users` / mutations

### DIFF
--- a/src/pages/users/[userId].page.tsx
+++ b/src/pages/users/[userId].page.tsx
@@ -1,0 +1,29 @@
+import { useRouter } from "next/router"
+import { useMetaphysics } from "system/artsy-next-auth"
+import { Text } from "@artsy/palette"
+
+const User: React.FC = () => {
+  const router = useRouter()
+  const { userId } = router.query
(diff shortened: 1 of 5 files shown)


src/pages/users/[userId].page.tsx
```diff
@@ -0,0 +1,49 @@
...
+  )
+}
+
+export const getServerSideProps: GetServerSideProps = async (ctx) => {
+  const props = await fetchRelayData({
+    query: graphql`
+      query UserIdQuery($userId: String!) {
+        me {
+          name
+        }
```
Comment (anandaroop) [inline src/pages/users/[userId].page.tsx:30]: Used?

anandaroop (approved):
LGTM, just one q below

src/pages/users/[userId].page.tsx
Comment (jonallured) [reply src/pages/users/[userId].page.tsx:30]: Yeah I was confused by this too - I just sorta copy/pasted from the artist example I saw.

PR #30, opened 2022-03-29
feat(user): Add basic user form and mutations

damassi:
This does a few things:
- Stubs out a basic user form via formik
- Sets up some form patterns 
- Adds a mutation 
- Adds some patterns for using mutations 
- Adds script for syncing MP localhost schema 

There's a weird issue with actually submitting the mutation -- it says that the token has expired. But that shouldn't hold up this PR


src/pages/users/UserForm.tsx
```diff
@@ -0,0 +1,54 @@
...
+          value={values.email}
+          onChange={handleChange}
+          onBlur={handleBlur}
+          error={touched.email && errors.email}
+          required
+        />
+      </Column>
+
+      {/* <Column span={6}>
+        <CountrySelect
```
damassi, line 37: Leaving this here for later since we have lots of selects

jonallured (commented):
Oooo laying down lots of nice patterns I like it! I'd be curious to know about test coverage. Should there be jest tests here? What about Cypress tests. I don't think it should be blocking but it's on my mind regardless.

src/hooks/useMutation.ts
```diff
@@ -0,0 +1,85 @@
...
+      },
+      rejectIf: res => {
+        return !res.updateUser?.clientMutationId
+      },
+    })
+  }
+ *
+ *
+ */
+export const useMutation = <T extends MutationParameters>({
```
jonallured, line 49: Nice example comment!! 😍

src/pages/users/[userId].page.tsx
```diff
@@ -2,33 +2,68 @@ import { GetServerSideProps } from "next"
...
   )
 }
 
 export const getServerSideProps: GetServerSideProps = async (ctx) => {
   const props = await fetchRelayData({
     query: graphql`
       query UserIdQuery($userId: String!) {
-        me {
-          name
-        }
```
Comment (jonallured) [inline src/pages/users/[userId].page.tsx:30]: It might be handy here to pull out roles - the first thing we could do with them is conditionally disable the email field. Only admins with `customer_support` in their roles can change that one!

damassi:
>  I'd be curious to know about test coverage

Things are still a bit up in the air regarding relay's SWC compiler and graphql tags in jest. I plan on looking at that in a bit, but for now i think this is mostly just a pattern pass. Once we fill out more details I'd be interested in adding Jest, but probably not during hackathon.

src/pages/users/useUserFormContext.tsx
```diff
@@ -0,0 +1,53 @@
...
+  receiveOutbidNotification: boolean
+  receivePromotionNotification: boolean
+  receivePurchaseNotification: boolean
+  receiveSaleOpeningClosingNotification: boolean
+  shareFollows: boolean
+  */
+}
+
+export const userValidationSchema = Yup.object({
+  email: Yup.string().required("Email is required"),
```
jonallured, line 47: Yup can get fancy by examining the `context` that's injected. I remember using that a lot on the Artwork Form but if memory serves I think @pepopowitz ended up changing course - can you remind me where you landing on that Steve?

damassi: (Note that I don't think we should be thinking about optimizing for performance at this stage)

src/pages/users/[userId].page.tsx
Comment (jonallured) [reply src/pages/users/[userId].page.tsx:30]: Ugh I'm just now realizing that roles are not at this level - they come from the token decode step.

PR #33, opened 2022-03-29
chore(prettier): ignore generated folder

damassi:
I keep noticing weird prettier faffing on our generated queries on git pull, so this ignores that directory.


jonallured:
Yeah this was super annoying thank!!

PR #34, opened 2022-03-29
feat(system): add SystemContext for system-level functions

damassi:
This adds a system context, similar to force, where you can pull out commonly used things such as `user` and `relayEnvironment`, etc. I expect this to grow in time. 

This also fixes the missing headers for mutations; however this now yields a new error which I suspect is a MP level thing, or personal access thing (which is good!):

Screen Shot 2022-03-29 at 2 30 20 PM


jonallured (approved):
LGTM!! 🎉

PR #39, opened 2022-03-30
chore: remove unused API endpoint

joeyAghion:
This status API was added (https://github.com/artsy/forque/pull/5) for kubernetes's readiness probe, but didn't end up working compatibly with the reorg in https://github.com/artsy/forque/pull/6. Instead, we added a "static" status route in https://github.com/artsy/forque/pull/11, which I think is actually fine for readiness probe purposes (since it's not being requested through a CDN or any other caching server). This PR just removes the original status API.


anandaroop:
Closing in favor of #50

PR #52, opened 2022-03-31
feat(user): Add pagination component and connection examples

damassi:
This adds a new `ListPagination` component and wires up some new patterns for iterating through lists of data.

Also upgrades React to 18 and updates README with a note on new React 18 patterns.


src/pages/users/components/UsersTable.tsx
```diff
@@ -0,0 +1,87 @@
...
+              internalID
+              email
+              name
+            }
+          }
+        }
+      }
+    `,
+    viewer
+  )
```
dzucconi, line 48: So much nicer than the PaginationContainers

src/pages/users/components/UsersTable.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+interface UsersTableProps {
+  viewer: UsersTable_viewer$key
+}
+
+export const UsersTable: React.FC<UsersTableProps> = ({ viewer }) => {
+  const [data, refetch] = useRefetchableFragment(
+    graphql`
+      fragment UsersTable_viewer on Viewer
+      @refetchable(queryName: "UsersTableQuery")
```
damassi, line 17: Note that here, rather than redefining a query that we use to refetch, we can construct it dynamically. This name just has to be unique, and the pattern i've adopted is `ComponentNameQuery`.

src/components/ListPagination.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { Pagination } from "@artsy/palette"
+import { ListPagination_pageCursors$key } from "__generated__/ListPagination_pageCursors.graphql"
+import { startTransition } from "react"
+
+const REFETCH_PAGE_SIZE = 20
+
+interface PaginationProps {
+  pageCursors: ListPagination_pageCursors$key
+  pageInfo: { hasNextPage: boolean; endCursor: string | null }
+  relayRefetch: RefetchFnDynamic<any, any>
```
damassi, line 11: This is a TODO.

At some point we'll want to update this interface with a generic to pass into the refetch. We should get types here, and can, as the `@refetchable` [directive here](https://github.com/artsy/forque/pull/52/files#r840003091) will spit out a type

PR #54, opened 2022-03-31
feat(table): Add <Table> component and wire up users

damassi:
This adds a new `<Table>` component and wires it up to the /users route.

For this I used [react-table](https://react-table.tanstack.com/) which is a cool table building toolchain, rather than a table component, which allows us to keep full control of everything including styles. It does a lot and I think it will scale out well to all of our needs, whatever they may be. 

table

The wrapper `<Table />` component that I wrote around it looks like so:

```tsx
 <Table
  columns={[
    {
      Header: "Name",
      accessor: "name", // accessor is the "key" in the data
    },
    {
      Header: "Email",
      accessor: "email",
    },
    {
      Header: "ID",
      accessor: "internalID",
    },
  ]}
  data={users}
  onRowClick={(row) => {
    router.push(`/users/${row.values.internalID}`)
  }}
/>
```

As things expand I imagine this API will grow.


src/components/Table.tsx
```diff
@@ -0,0 +1,74 @@
+import { Box, Text, THEME_V3 } from "@artsy/palette"
+import { useMemo } from "react"
+import { Row, useTable } from "react-table"
+import styled from "styled-components"
+
+interface TableProps {
+  columns: any[]
```
damassi, line 7: Need to figure out generic types

src/components/Table.tsx
```diff
@@ -0,0 +1,74 @@
...
+  data: any[]
+  onRowClick: (row: Row) => void
+}
+
+export const Table: React.FC<TableProps> = ({ columns, data, onRowClick }) => {
+  const columnsMemo = useMemo(() => columns, [columns])
+  const dataMemo = useMemo(() => data, [data])
```
dzucconi, line 14: Why do you have to memo these props?

src/components/Table.tsx
dzucconi: Something like:

```tsx
interface TableProps<T, U> {
  columns: T[]
  data: U[]
}

export const Table = <T, U>({ columns, data }: TableProps<T, U>) => {
  // ...
```

dzucconi (approved):

src/components/Table.tsx
damassi: This is what the docs said was required

src/components/Table.tsx
damassi: 👍 will address next week

PR #55, opened 2022-04-01
Unmergeable but successfully integrated oauth

joeyAghion:
Thanks for the pairs!

This PR demonstrates oauth via Gravity using the `next-auth` library.  I had to comment out a lot of session-related code and bypass type-checking and other lint-ing to commit this (so it's not mergeable), but a new `/authenticated` route successfully shows Gravity auth status and offers a button to sign in or out:

Screen Shot 2022-04-01 at 12 17 44 AM

⬇️ 

Screen Shot 2022-04-01 at 12 17 58 AM

⬇️ 

Screen Shot 2022-04-01 at 12 18 15 AM

⬇️ 

Screen Shot 2022-04-01 at 12 18 51 AM

We had to customize the basic oauth config with:
* A special parameter to Gravity's token endpoint (added in https://github.com/artsy/gravity/pull/15183) causes it to return a `200` rather than `201` status code.
* The userinfo request had to be customized to pass the `X-Access-Token: ...` header instead of the more conventional `Authorization: Bearer ...`.

I think the next step would be reconnecting the layout and relay to the session provided by `next-auth`.


src/pages/api/auth/[...nextauth].page.ts
```diff
@@ -0,0 +1,39 @@
...
+    {
+      id: "artsy",
+      clientId: process.env.CLIENT_APPLICATION_ID,
+      clientSecret: process.env.CLIENT_APPLICATION_SECRET,
+      name: "Artsy",
+      type: "oauth",
+      authorization: `${process.env.GRAVITY_URL}/oauth2/authorize`,
+      token: {
+        url: `${process.env.GRAVITY_URL}/oauth2/access_token?on_success=200`,
+        params: { on_success: 200 }
```
Comment (joeyAghion) [inline src/pages/api/auth/[...nextauth].page.ts:15]: Strangely, this only worked when the parameter was _both_ here and in the URL above.

anandaroop:
Exciting to see the Force login screen when signing in to Forque 🎉 

Noting that I got a warning locally when I went to `yarn install` — `next-auth` [requires](https://github.com/nextauthjs/next-auth/blob/e936c51575b5a19aa604374ad40746db83cc07e0/package.json#L50) Node v16.13+ so we'll probably want to update [.nvmrc](https://github.com/artsy/forque/blob/main/.nvmrc) as well

src/pages/api/auth/[...nextauth].page.ts
Comment (mzikherman) [reply src/pages/api/auth/[...nextauth].page.ts:15]: Weird! We should look as to why that seems to be needed in the lib (or can simply revert to `token: ...` usage)

anandaroop:
Closing in favor of #60

PR #58, opened 2022-04-01
refactor(nav): converts nav to palette

dzucconi:
I don't really like the way this looks — our color scheme really wasn't designed for these inverted use-cases — but this converts the nav to Palette and adds active link highlighting.

### DIFF
--- a/src/components/__tests__/GlobalNav.spec.tsx
+++ b/src/components/__tests__/GlobalNav.spec.tsx
@@ -3,6 +3,8 @@ import { render, screen } from "@testing-library/react"
 
 import { GlobalNav } from "../GlobalNav"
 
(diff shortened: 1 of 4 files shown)


src/components/__tests__/GlobalNav.spec.tsx
```diff
@@ -3,6 +3,8 @@ import { render, screen } from "@testing-library/react"
 
 import { GlobalNav } from "../GlobalNav"
 
+jest.mock("next/router", () => ({ useRouter: () => ({ pathname: "/" }) }))
```
anandaroop, line 6: This is so nice and succinct. I wish I could always remember the 50 different ways to mock things in Jest.

anandaroop (approved):
Maybe dropdown menu components would be more suitable eventually?

Looks good for now though 👍🏽 thanks for the palettizing.

PR #59, opened 2022-04-02
feat(autocompletes): add standard autocompletes for artist and user

anandaroop:
The goal here is to sketch out some reusable best-practice autocompletes components that will:

- have a simple interface, e.g.

  ```tsx
  <ArtistAutocomplete onSelect={artist => doSomethingWith(artist)} />
  ```

-  include all the disambiguating info needed to correctly select an item, which is especially important for artists

- set up patterns that can be repeated for other frequently accessed content types (artwork, partner, show, etc) 

This screencap below shows the Artist version. (There is also a **`UserAutocomplete`**, but I'll refrain from screencapping due to all the PII)

https://user-images.githubusercontent.com/140521/177420421-a6434b4b-af19-4590-a61b-08fed2f49a9b.mov

## REST vs Relay

This was very straightforward to implement with a fetch to Gravity's REST endpoints, but I thought it wise to at least _try_ and implement this in Relay. [Not much luck so far](https://artsy.slack.com/archives/C0376CFU527/p1657036417620809), but if we figure that out we can come back and make another pass at these. Since they are very self-contained, shouldn't matter to the call sites.

### TODO

- [x] Rough in a design
- [x] Wire it to live data
- [x] Figure out how to test
- ~Add Storybook entries~ 
  - not straightforward bc access tokens. Added a demo page at `/auto` for now
- [ ] Switch data fetching to Relay?
  - [x] paired with Damon on this, inconclusively so far

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
-const { withSentryConfig } = require('@sentry/nextjs');
+const { withSentryConfig } = require("@sentry/nextjs")
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
@@ -23,9 +23,16 @@ const nextConfig = {
   sentry: {
     disableServerWebpackPlugin: true,
     disableClientWebpackPlugin: true,
-  }
+  },
+  images: {
+    domains: getImageDomains(),
+  },
+}
+
+function getImageDomains() {
+  return (process.env.ALLOWED_IMAGE_DOMAINS || "").split(/\s*,\s*/)
 }
 
 // Make sure adding Sentry options is the last code to run before exporting, to
 // ensure that your source maps include changes from all other Webpack plugins
-module.exports = withSentryConfig(nextConfig);
+module.exports = withSentryConfig(nextConfig)
--- a/src/components/autocomplete/ArtistAutocompleteOption.tsx
+++ b/src/components/autocomplete/ArtistAutocompleteOption.tsx
@@ -1,0 +1,56 @@
+import Image from "next/image"
+import compact from "lodash/compact"
+import {
+  AutocompleteInputOptionLabelProps,
+  Box,
+  Flex,
+  Text,
+} from "@artsy/palette"
+
+export interface Props extends AutocompleteInputOptionLabelProps {
+  imageUrl: string
+  isTargetSupply: boolean
+  nationality: string
+  publishedArtworksCount: number
+  targetSupplyPriority: number
+  years: string
+}
+
+export const ArtistAutocompleteOption: React.FC<Props> = (props) => {
+  const {
+    imageUrl,
+    nationality,
+    publishedArtworksCount,
+    targetSupplyPriority,
+    text,
+    years,
+  } = props
+  const priority = targetSupplyPriority && `P${targetSupplyPriority}`
+  const works = publishedArtworksCount && `${publishedArtworksCount} works`
+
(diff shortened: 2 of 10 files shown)


src/components/autocomplete/ArtistAutocompleteOption.tsx
```diff
@@ -0,0 +1,58 @@
...
+            layout="fill"
+          />
+        )}
+      </Box>
+      <Box flex={1} p={1}>
+        <Text overflowEllipsis fontWeight={"bold"}>
+          {text}
+        </Text>
+
+        <Text lineHeight={1.5} overflowEllipsis textColor="black60">
```
dzucconi, line 52: Just use one of the Text variants https://palette-storybook.artsy.net/?path=/story/components-text--variants — rather than customizing the typography.

dzucconi (approved):

next.config.js
```diff
@@ -23,9 +23,12 @@ const nextConfig = {
   sentry: {
     disableServerWebpackPlugin: true,
     disableClientWebpackPlugin: true,
-  }
+  },
+  images: {
+    domains: ["d32dm0rphc51dk.cloudfront.net", "d2v80f5yrouhh2.cloudfront.net"],
+  },
```
artsyjian, line 29: Interesting. Is this like an allow-list of domains? If it's only these two domains, I wonder if it's better to fish them out of Env?

src/components/autocomplete/ArtistAutocompleteOption.tsx
anandaroop: Ah yeah, those overrides were mostly necessary. Updated.

next.config.js
anandaroop: Yep, when you use Next's `<Image />` component, it will get run through Next's image pipeline, and thus [needs this allow-list to protect](https://nextjs.org/docs/basic-features/image-optimization#domains) from abuse. 

Sure, we could go the .env route. Since these are not secrets, I assume they can just go in `.env.development` — I'll add them there.

dzucconi: Personally not really a fan of Next's image handling, and if we were using the Relay query we would use our own on-demand resizing (for now this is fine, of course).

src/components/autocomplete/UserAutocompleteOption.tsx
```diff
@@ -0,0 +1,45 @@
...
+        >
+          {email}
+          {isIdentityVerified && (
+            <span
+              data-testid="identity-verified-badge"
+              style={{ position: "relative" }}
+            >
+              <Box position="absolute" top="0" left="0.25em">
+                <VerifiedIcon fill="green100" height={20} width={20} />
+              </Box>
```
anandaroop, line 38: Forgot to call out that the user autocomplete will add a badge for any identity-verified accounts.

Screen Shot 2022-07-06 at 1 07 34 PM

next.config.js
artsyjian: Ah cool. Oh yeah, if the domains are constant across environments, then having them here seems fine to me too.

anandaroop:
## Migration

Setting [these](https://github.com/artsy/forque/pull/59/commits/6af5b15916a3e82826315054e458e253582c134a#diff-da4c41d59c967338247b2f0ea6d845ab05171f942fd9521b96d3304202185db9R15) in stage and prod now so that they'll be ready to go

anandaroop:
Alright time to 🚢

PR #60, opened 2022-04-07
feat(auth): replace homegrown auth code with Gravity OAuth, via NextAuth.js

anandaroop:
A big step towards a more standardized auth flow:

- Removes the homegrown authentication code that has been part of this repo since the original spike

- Replaces it with the community-created library [NextAuth.js](https://next-auth.js.org), which allows for custom OAuth providers

- Configures a custom `artsy` OAuth provider that delegates to Gravity (which also led to some [adjustments to Gravity](https://github.com/artsy/gravity/pull/15183))

- Wires up the resulting user session and token to the data fetching libraries

- Requires a valid session in order to render just about anything

This is probably enough for one PR, but some likely follow-ups would be:

- Testing (see [comment](https://artsy.slack.com/archives/C0376CFU527/p1649264539333269?thread_ts=1649164933.690199&cid=C0376CFU527))

- Role-based control so that only `admin` users have access

## When initially logging in

You will be redirected to our common login form in Force

https://user-images.githubusercontent.com/140521/162117815-64715778-2b17-4d2d-898a-a787e7d6fd65.mov

## When you already have a valid Gravity session

You will skip the login form, and get a new Forque session immediately

https://user-images.githubusercontent.com/140521/162117856-11be89e7-39e7-4649-a4af-234a4f20c68b.mov

## When you force a logout from Forque as well as from Gravity

You'll get kicked back to the common login form in Force

https://user-images.githubusercontent.com/140521/162117905-1fa432e8-1de0-41bf-a72b-c3d298677f15.mov

## After you've logged in

Data-fetching continues to work, as before

https://user-images.githubusercontent.com/140521/162118226-f9df5fd7-ab3f-4b23-b357-17b885dedb7c.mov


src/components/Layout.tsx
```diff
@@ -1,32 +1,51 @@
...
 
-            <Box position="fixed" zIndex={10} bottom={1} right={1} width={400}>
-              <Toasts />
-            </Box>
-          </main>
-        </>
-      ) : (
-        <LoginForm />
-      )}
+      <main className="container mx-auto p-2 sm:py-4 text-center">
```
damassi, line 46: ditto

src/pages/_app.page.tsx
```diff
@@ -23,32 +22,33 @@ export default function MyApp({ Component, pageProps }: AppProps) {
...
-              </Layout>
-            </ErrorBoundary>
-          </ToastsProvider>
-        </Theme>
-      </RelayEnvironmentProvider>
-    </SystemContextProvider>
+    <SessionProvider session={pageProps.session}>
+      <SystemContextProvider
+        relayEnvironment={environment}
+        user={pageProps.systemUser}
```
damassi, line 28: There's a couple things we can do here: 

1. We'll want to update the system context to receive `session` rather than `user` or eliminate this passed prop completely and instead set `session` via the `useSession` hook internally within `SystemContextProvider`. 
1. Remove `session` from the `SystemContext` completely since we can grab it via the `useSession` hook from the `SessionProvider`
1. (Ideally): Remove the `SystemContext` completely since we can grab `session` from `useSession` hook and the relay environment from the `useRelayEnvironment` [hook](https://relay.dev/docs/api-reference/use-relay-environment/)

src/pages/_app.page.tsx
```diff
@@ -23,32 +22,33 @@ export default function MyApp({ Component, pageProps }: AppProps) {
...
 
 MyApp.getInitialProps = async (appContext: AppContext) => {
+  const session = await getSession(appContext.ctx)
   const appProps = await App.getInitialProps(appContext)
-  const systemUser = await getUserFromCookie(
-    appContext.ctx.req as NextApiRequest
-  )
-  appProps.pageProps.systemUser = systemUser
+  appProps.pageProps.session = session
+  appProps.pageProps.systemUser = session?.user
```
damassi, line 52: This can go away since `user` is already on `session` and we can get it from there

src/system/artsy-next-auth/index.ts
```diff
@@ -1,8 +1,7 @@
-// user authentication and session
-export { LoginForm } from "./auth/login-form"
-export { handleAuth } from "./auth/api"
-export { useUser } from "./hooks/user"
+import type { User } from "next-auth"
```
damassi, line 1: Can we remove the `artsy-next-auth` folder and perhaps rename/move the `useGravity.ts` and `useMetaphysics.ts` hooks to `src/hooks`?

src/system/relay/fetchRelayData.ts
```diff
@@ -1,12 +1,13 @@
...
+import { getSession } from "next-auth/react"
 import {
   CacheConfig,
   fetchQuery,
   FetchQueryFetchPolicy,
   GraphQLTaggedNode,
   OperationType,
 } from "relay-runtime"
-import { getUserFromCookie } from "system/artsy-next-auth/auth/user"
+import { UserWithAccessToken } from "system/artsy-next-auth"
```
damassi, line 10: This can live in `src/system/index.ts`

src/system/artsy-next-auth/index.ts
```diff
@@ -1,8 +1,7 @@
...
 
 // data fetching
 export { useGravity } from "./hooks/gravity"
 export { useMetaphysics } from "./hooks/metaphysics"
+
+export type UserWithAccessToken = User & { accessToken: string }
```
damassi, line 7: With the other files moved out and this folder gone, this straggler type can live in `src/system/index.ts`

damassi (approved):
A handful of minor comments but otherwise this looks really sweet. Great work on the approach 💯

src/components/Layout.tsx
```diff
@@ -1,32 +1,51 @@
...
 interface LayoutProps {
-  user: UserSessionData
+  user: UserWithAccessToken
 }
 
 export const Layout: React.FC<LayoutProps> = ({ children, user }) => {
+  return user ? (
+    <AuthorizedLayout user={user}>{children}</AuthorizedLayout>
+  ) : (
+    <UnauthorizedLayout />
```
damassi, line 14: Would it make sense to remove some friction and do something like this instead of adding an additional UI interaction:

```tsx
const Layout = (...) => {
  useEffect(() => {
    if (!user) {
      signIn("artsy")
    }
  }, [user])
  
  if (!user) {
    return null
  }
  
  return (
    <Layout>{children}</Layout>
  )
}
```
Then we shouldn't need the `Authorized`/`Unauthorized` divide. 

(Also note that we shouldn't need to pass `user` into layout since we can grab it from the `useSession` hook.)

damassi: As a related thought, now that we've got this sweet oauth flow through artsy.net we should update Force's /login route to accept a few more query params for setting an additional title that mentions the referrer, just so the user knows where they're coming from -- that or output the redirectUrl. That could reduce some UX confusion down the line.

joeyAghion:
I started on the role-based authorization here: https://github.com/artsy/forque/commit/65f00045d487d3429ac9a2e1266d50378266c912

That should probably be a later PR, and needs review for type-safety and idiomacy.

src/system/artsy-next-auth/index.ts
anandaroop: Good call 33131ca

src/system/artsy-next-auth/index.ts
anandaroop: Good call — hadn't realized `src/hooks` was already a thing! 3d4602a0d685b0986007934494d1e1468a6c3f9c

src/pages/_app.page.tsx
anandaroop: Yep 2db4533

src/pages/_app.page.tsx
anandaroop: Everything except # 3 in 2db4533 and 9ce05ce — might leave that last bit for you

anandaroop:
Think I covered everything except one small piece of the Relay config. 

Giving this a merge now 🚀

src/pages/_app.page.tsx
damassi: Will take care of it!

src/hooks/index.ts
```diff
@@ -0,0 +1,6 @@
+export { useClipboard } from "./useClipboard"
+export { useExtractNodes } from "./useExtractNodes"
+export { useGravity } from "./useGravity"
+export { useMetaphysics } from "./useMetaphysics"
+export { useMutation } from "./useMutation"
+export { useSystemContext } from "./useSystemContext"
```
damassi, line 6: 👍

PR #63, opened 2022-04-08
feat: introduce My Collection artwork transfer

olerichter00:
Addresses [CX-2383](https://artsyproduct.atlassian.net/browse/CX-2383)

## Description

This PR adds a new "My Collection" page with functionality to transfer artworks from one user's collection to another user's collection.


src/utils/validateEmail.ts
```diff
@@ -0,0 +1,7 @@
+export const validateEmail = (email: string) => {
+  return email
+    .toLowerCase()
+    .match(
+      /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+    )
+}
```
dzucconi, line 7: If we're using Formik and Yup — should just use Yup's email validation (or just `type="email"` is frequently good enough)

src/pages/my-collection/components/UserCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+}> = ({ fieldName, fieldValue, loading }) => {
+  return (
+    <>
+      <Text color="black60" variant="xs">
+        {fieldName}
+      </Text>
+      {loading ? (
+        <SkeletonText>{fieldName}</SkeletonText>
+      ) : (
+        <Text color="black">{fieldValue ?? "/"}</Text>
```
dzucconi, line 96: `Text` should always have a `variant` (reminds me to change the default 😬 )

src/pages/my-collection/components/UserCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+      const response = await fetchUserByEmail(email)
+
+      setUser(response?.user)
+    } catch (error) {
+      setError(true)
+      console.error(error)
+    } finally {
+      setLoading(false)
+    }
+  }
```
dzucconi, line 34: Wouldn't Relay hooks do this all for you?

olerichter00: I've tried it but I'm expecting the query to fail if the user is not found and ErrorBoundary wasn't catching the error correctly. Any ideas?

dzucconi: I wouldn't expect `ErrorBoundary` to catch it because it's not an exception. You could always just pull the error off and if it exists throw it. `if (error) throw error`

olerichter00: Ah, the problem was that the UI was crashing even though I wrapped the component with ErrorBoundary.

I'll give it another try. In theory it should work 🤔

src/pages/my-collection/components/MyCollectionTransfer.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <Text mb={4}>
+        {`Transfer artwork from one user's collection to another user's collection.`}
+      </Text>
+      <GridColumns>
+        <Column span={5} minHeight={[0, 300]}>
+          <Input
+            title="From"
+            placeholder="Please enter an email address"
+            onChange={(e) => setEmailFrom(e.target.value)}
+          />
```
damassi, line 74: (blocking) We should be using Formik

src/pages/my-collection/queries/useFetchUserByEmail.ts
```diff
@@ -0,0 +1,22 @@
+import { fetchQuery, graphql } from "relay-runtime"
+import { useEnvironment } from "system/relay"
+import { useFetchUserByEmailQuery } from "__generated__/useFetchUserByEmailQuery.graphql"
+
+export const useFetchUserByEmail = () => {
+  const environment = useEnvironment()
```
damassi, line 6: This will all change slightly once @anandaroop merges https://github.com/artsy/forque/pull/60, but we should use relay's own [`useRelayEnvironment`](https://relay.dev/docs/api-reference/use-relay-environment/) hook

src/pages/my-collection/components/UserCard.tsx
damassi: So this hook doesn't work? https://relay.dev/docs/api-reference/use-lazy-load-query/

olerichter00: Exactly, I tried to use `[use-lazy-load-query/](https://relay.dev/docs/api-reference/use-lazy-load-query/)` and I'm getting this error which `ErrorBoundary` didn't catch:

```
RRNLRequestError: Relay request for `useFetchUserByEmailQuery` failed by the following reasons:

1. Cannot return null for non-nullable field User.internalID.
     user(email: $email) {
     ^^^
    at createRequestError (createRequestError.js?9316:110:1)
    at eval (fetchWithMiddleware.js?90dc:93:1)
```

dzucconi: OK well that's definitely an issue in that there's apparently a user without an internal ID? I don't see how the method of query would change this though.

damassi: Interesting, i bet this is a framework setup thing. I can investigate the error boundary thing today.

olerichter00: Could the issue be that Error boundaries only catch the error if the error happens in the React render lifecycle and it's is not the case if a request error happens?

dzucconi: Like this error is clearly a problem but. It shouldn't be the case that a failed fetch would get caught by the ErrorBoundary. The ErrorBoundary is only going to catch React errors. Data fetching errors typically don't trigger unless you explicitly raise it by throwing (handling data error states is a normal part of building a component, not an exceptional state).

damassi: @dzucconi - Am I misreading this from the docs? 

> useLazyLoadQuery will use [Suspense](https://relay.dev/docs/guided-tour/rendering/loading-states/) to allow developers to render loading states using Suspense boundaries, and will throw errors if network errors occur, which can be caught and rendered with Error Boundaries.

Doesn't this seem to suggest that a GraphQL error of the sort described above: 
> Cannot return null for non-nullable field User.internalID.
would be caught here, and then handled at the error boundary level? This suspense-style thinking is new to me so I'm certain I'm misunderstanding something somewhere.

dzucconi: OH, yeah I'm just basing this off of working with Apollo and Urql hooks. Interesting. I guess this is a new thing.

olerichter00: It's now working with `useLazyLoadQuery`, https://github.com/artsy/forque/pull/70 fixed the issue.

PR #69, opened 2022-04-08
feat(auth): require relevant roles for access to this application

joeyAghion:
Restrict access to roles that are able to use these administrative functions in 2 ways:

* Consider users with _none_ of the applicable roles unauthorized from the whole application, showing them the built-in access denied page. Looks like:
Screen Shot 2022-04-07 at 12 58 54 PM

* Only render navigation to functions that are accessible to each role. This doesn't truly guard anything, obviously. Similar checks should be made on the pages themselves, and the back-ends are the ultimate authority and enforcer of which functions are allowed for which roles. These explicit checks should probably be replaced by shared helpers, in this PR (if you can advise about where that should go) or later. **Forque's authorization checks are a UX convenience only.**

### DIFF
--- a/src/pages/api/auth/[...nextauth].page.ts
+++ b/src/pages/api/auth/[...nextauth].page.ts
@@ -1,17 +1,27 @@
 import NextAuth from "next-auth"
 import { UserinfoEndpointHandler } from "next-auth/providers"
 
+const authorizedRoles = ["admin", "customer_support", "team"] // all roles relevant to this app
+
 export default NextAuth({
   callbacks: {
-    jwt: async ({ token, account }) => {
+    // only allow those with relevant roles to sign in
+    signIn: async ({ profile }) => {
+      const userRoles = (profile.roles as string[]) || []
+      return authorizedRoles.some((r) => userRoles.includes(r))
+    },
+    jwt: async ({ token, user, account }) => {
       if (account) {
         token.access_token = account.access_token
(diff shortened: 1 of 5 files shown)


src/pages/api/auth/[...nextauth].page.ts
```diff
@@ -1,17 +1,27 @@
...
 
+const authorizedRoles = ["admin", "customer_support", "team"] // all roles relevant to this app
+
 export default NextAuth({
   callbacks: {
-    jwt: async ({ token, account }) => {
+    // only allow those with relevant roles to sign in
+    signIn: async ({ profile }) => {
+      const userRoles = (profile.roles as string[]) || []
+      return authorizedRoles.some((r) => userRoles.includes(r))
```
Comment (anandaroop) [inline src/pages/api/auth/[...nextauth].page.ts:11]: Nice, so this boolean return value determines whether or not they see the access denied page 👍🏽

anandaroop (approved):
LGTM!

I tested locally with an internal user with sufficient roles, and a non-artsy user.  Works as expected 👍🏽 

I _also_ inadvertently tested the case of an internal user lacking the necessary roles, because when I booted the app I still had a valid session, but my user lacked any `roles`. Also worked as expected 👍🏽

anandaroop:
As soon as some CONTRIBUTING.md docs take place, would be good to recommend this practice of gating at the frontend.

PR #70, opened 2022-04-08
chore(relay): dont throw on errors

damassi:
This fixes our relay layer to not throw on errors, so that we can properly handle them within the app. Also cleans up the nav a bit.

### DIFF
--- a/src/system/relay/middleware.ts
+++ b/src/system/relay/middleware.ts
@@ -1,6 +1,7 @@
 import getConfig from "next/config"
 import {
   cacheMiddleware,
+  errorMiddleware,
   Headers,
   loggerMiddleware,
   Middleware,
@@ -9,6 +10,9 @@ import {
 import type { UserWithAccessToken } from "system"
 
 const { publicRuntimeConfig } = getConfig()
+
+const enableLogging =
+  process.env.NODE_ENV === "development" && typeof window !== "undefined"
 
 export const getRelayMiddleware = (
   user?: UserWithAccessToken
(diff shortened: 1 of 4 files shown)


src/system/relay/middleware.ts
```diff
@@ -20,18 +24,18 @@ export const getRelayMiddleware = (
...
     urlMiddleware({
       url: publicRuntimeConfig.NEXT_PUBLIC_METAPHYSICS_URL! + "/v2",
       headers: authenticatedHeaders,
     }),
-    loggerMiddleware({
-      logger: (...args: any) =>
-        typeof window !== "undefined" ? console.log(...args) : undefined,
-    }),
+    enableLogging && loggerMiddleware(),
+    enableLogging && errorMiddleware({ disableServerMiddlewareTip: true }),
```
dzucconi, line 33: 👍

dzucconi (approved):

src/system/relay/setupEnvironment.ts
```diff
@@ -32,7 +32,7 @@ export function setupEnvironment({
 
 function createEnvironment(user?: UserWithAccessToken) {
   return new Environment({
-    network: new RelayNetworkLayer(getRelayMiddleware(user)),
+    network: new RelayNetworkLayer(getRelayMiddleware(user), { noThrow: true }),
```
damassi, line 35: This is the main change, but note that this might have implications elsewhere related to Suspense. Need to experiment a bit more in practice to see how it shakes out.

PR #75, opened 2022-04-13
feat: Admin can manage sale profile data for a user

jonallured:
This PR lands the management of a nested user model - the `SaleProfile`. This is a layer on top of our user data where we can override certain details if we want. In order to support this editing I added a query/mutation pair here:

https://github.com/artsy/metaphysics/pull/3952

The query is under `user` and the mutation is root-level. So yeah these fields are wired up!

Screen Shot 2022-04-13 at 4 09 03 PM


dzucconi (approved):
😗👌

damassi (approved):
Nice and simple!

PR #82, opened 2022-04-14
chore: remove old internal ingress declarations #migration (PLATFORM-4093)

joeyAghion:
This follows up https://github.com/artsy/forque/pull/76 by removing the internal ingress resources that are no longer used.

**Post-deploy, these resources will need to be manually deleted from staging and production clusters because that won't happen automatically.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@ A home for admin apps
 ## Meta
 
 - **State:** development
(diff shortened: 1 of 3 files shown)


ovasdi (approved):
🦖

ovasdi:
@joeyAghion deleted both (_prod, stg_) internal ingresses.

PR #84, opened 2022-04-14
Streamline authorization

joeyAghion:
This PR does 2 [related] things, so it might be easiest to review the commits separately.

1. Avoid relying on the legacy, overly-broad `admin` role for access to any tools. This is just a small step, and APIs will ultimately need to narrow their access as well. At least this limits the new functionality granted to `admin`s and decreases the need for that role.

2. Add an `isPermitted` helper so that authorization rules can be declared clearly in just one location, then referenced everywhere. Tell me what you think of the interface!

It bears repeating that these authorization rules are only a UX convenience, and ultimately enforcement must be at the API(s).

### DIFF
--- a/src/components/GlobalNav.tsx
+++ b/src/components/GlobalNav.tsx
@@ -5,6 +5,7 @@ import Link from "next/link"
 import { useRouter } from "next/router"
 import { FC } from "react"
 import styled, { css } from "styled-components"
+import { isPermitted, Action } from "system"
 import type { UserWithAccessToken } from "system"
 
 interface GlobalNavProps {
@@ -31,16 +32,16 @@ export const GlobalNav: FC<GlobalNavProps> = ({ user }) => {
           // Logged in
           <>
             <Item href="/">Home</Item>
-            {user.roles.includes("customer_support") && (
+            {isPermitted(user, [Action.list], "users") && (
               <Item href="/users">Users</Item>
             )}
-            {user.roles.includes("metadata_admin") && (
+            {isPermitted(user, [Action.dedupe], "artists") && (
               <Item href="/artists/dedupe">Dedupe Artists</Item>
             )}
-            {user.roles.includes("customer_support") && (
+            {isPermitted(user, [Action.transfer], "my_collection") && (
               <Item href="/my-collection">My Collection</Item>
             )}
-            {user.roles.includes("team") && (
+            {isPermitted(user, [Action.list, Action.create], "uploads") && (
               <Item href="/uploads">Uploads</Item>
             )}
             <Item href="#" onClick={() => signOut()}>
--- a/src/components/GlobalNav.tsx
+++ b/src/components/GlobalNav.tsx
@@ -31,15 +31,13 @@ export const GlobalNav: FC<GlobalNavProps> = ({ user }) => {
           // Logged in
           <>
             <Item href="/">Home</Item>
-            {(user.roles.includes("admin") ||
-              user.roles.includes("customer_support")) && (
+            {user.roles.includes("customer_support") && (
               <Item href="/users">Users</Item>
             )}
-            {user.roles.includes("admin") && (
+            {user.roles.includes("metadata_admin") && (
               <Item href="/artists/dedupe">Dedupe Artists</Item>
             )}
-            {(user.roles.includes("admin") ||
-              user.roles.includes("customer_support")) && (
+            {user.roles.includes("customer_support") && (
               <Item href="/my-collection">My Collection</Item>
             )}
             {user.roles.includes("team") && (
--- a/src/system/index.ts
+++ b/src/system/index.ts
@@ -4,3 +4,50 @@ export type UserWithAccessToken = User & {
   accessToken: string
   roles: string[]
 }
+
+// all supported roles
+export enum Role {
+  customer_support = "customer_support",
+  metadata_admin = "metadata_admin",
+  team = "team",
+}
+
+export enum Action {
+  list,
+  dedupe,
+  create,
+  transfer,
+}
+
+// For each _domain_, a map of _actions_ to the authorized _roles_.
+const PERMISSIONS: Record<string, Record<string, Role[]>> = {
+  users: {
+    [Action.list]: [Role.customer_support],
+  },
+  artists: {
+    [Action.dedupe]: [Role.metadata_admin],
+    [Action.list]: [Role.metadata_admin],
+  },
+  my_collection: {
+    [Action.transfer]: [Role.customer_support],
+  },
+  uploads: {
+    [Action.list]: [Role.team],
+    [Action.create]: [Role.team],
+  },
+}
+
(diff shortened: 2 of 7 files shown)


src/system/index.ts
```diff
@@ -4,3 +4,42 @@ export type UserWithAccessToken = User & {
...
+  actions: string[],
+  object: keyof typeof PERMISSIONS
+) => {
+  return actions.some((action) => {
+    // @ts-ignore
+    const permittedRoles: string[] = PERMISSIONS[object][action] || []
+    return permittedRoles.some((permittedRole) =>
+      user.roles.includes(permittedRole)
+    )
+  })
```
joeyAghion, line 44: I couldn't figure out how to avoid the `@ts-ignore`. [Here's a playground](https://www.typescriptlang.org/play?#code/PTAEEMBtNBnBXADog9gJwC4FMAmo0qRawBQAxigHawagBKA8gDICiAyqALygDeJooMvBooAtljQB9BMnQYAXKABEQkeKkzUmJQBp+ocRnA5wRycdEBLSoqWHjp8OZxXKu-dnCjbn0UpIAviTkVDSgAAosdACyAJJsbLEMAHIc3HwCwhKwihkCoJCWNIoA2oysbAB0qhhiEtJIWhgAunoCAW0QmEUYObz6Aji4SFil5eyV9iZmFtatAwU9Y8wTU47OrvPtnRTQWGQYlqG5Cxho4NQAZhLLFdXCteoNsphboB36SJAoxn15AoViqAyisqr43gIyGgsKZRsDxmCYaI3kEgiFqLQsmguP18vhCMRSkpLigULplAAjcBoJTNQLBCgY0BFcISKwYbB4bgACnAByO1EUNDQ1gA5iVWqAUBSAFb7BSgADWWAAnihLqAMCrEFh1REonEEklUgBKLgAPlx+CwGHgaEoEH5oUqsDq3N5TsoZs4lv+BRtoB1aHZnLoBL6wrFEpxkRi8USKTYJWlcoOzRKfMOVGaoAAPrngXS8dbbfbA2zLBzcGGiLAXW7uUGQ9WCd7LVjKgRa5VrGRIPAhrBGxWqzga1gTSb9AEp2jGa6iJVvqLuSyR5zuSUlICMLSdMpqYcaLAlJOSEA) if you want to try.

src/components/GlobalNav.tsx
```diff
@@ -5,6 +5,7 @@ import Link from "next/link"
 import { useRouter } from "next/router"
 import { FC } from "react"
 import styled, { css } from "styled-components"
+import { isPermitted } from "../system"
```
anandaroop, line 8: #minor the project-relative import `from "system"` should work here.

src/components/GlobalNav.tsx
```diff
@@ -31,18 +32,16 @@ export const GlobalNav: FC<GlobalNavProps> = ({ user }) => {
...
+            {isPermitted(user, ["list"], "users") && (
               <Item href="/users">Users</Item>
             )}
-            {user.roles.includes("admin") && (
+            {isPermitted(user, ["dedupe"], "artists") && (
               <Item href="/artists/dedupe">Dedupe Artists</Item>
             )}
-            {(user.roles.includes("admin") ||
-              user.roles.includes("customer_support")) && (
+            {isPermitted(user, ["transfer"], "collections") && (
```
anandaroop, line 41: Should this be `my_collections` or something, to avoid the inevitable name collision with the _other_ collections aka marketing collections? 🙈

src/system/index.ts
```diff
@@ -4,3 +4,42 @@ export type UserWithAccessToken = User & {
...
+    create: [ROLES.team],
+  },
+}
+
+export const isPermitted = (
+  user: UserWithAccessToken,
+  actions: string[],
+  domain: keyof typeof PERMISSIONS
+) => {
+  return actions.some((action) => {
```
anandaroop, line 38: At a glance this says to me that if _any_ supplied action is permitted for that role, then the return value is `true`. Not sure if this is correct or intended, but specs would be very helpful to clarify this.

src/system/index.ts
anandaroop: I tried the TS brainteaser but don't have a quick answer.

src/components/GlobalNav.tsx
joeyAghion: I had that initially, and it worked when running the app, but the `GlobalNav` tests failed until I switched to this style. I still don't understand why!

src/components/GlobalNav.tsx
joeyAghion: Shoot you're right. I'm not crazy about that since "my" makes very little sense in this context. I guess I'll make the change anyway, since some disambiguation will be needed.

src/system/index.ts
joeyAghion: That is correct. Since these are only shared view helpers for convenience, my thinking was that if the authenticated user has access to _any_ of the specified functions, the return value should be `true`.

Specs are a very reasonable idea and I'll try that.

jonallured:
I'm curious about the intention here @joeyAghion - is this just a convenience for users so they don't see things in the nav that they don't have access to? If a user had book marked the merge artists page and then had it removed from their roles wouldn't that bookmark still work but then (assuming gravity is updated) they would just see errors about access? Should forque suppress more than just the nav?

joeyAghion:
@jonallured this is all just for UX, but I'd hope to apply the same checks in the nav and in the individual pages (and maybe even around particular _sections_ of pages, such as role-editing for users). I just wanted to iron out this API before adding it in more places.

src/system/index.ts
olerichter00: Adding a JSDoc comment for VS Code's intellisense could be helpful as well.

src/system/index.ts
```diff
@@ -4,3 +4,42 @@ export type UserWithAccessToken = User & {
...
+    // @ts-ignore
+    const permittedRoles: string[] = PERMISSIONS[domain][action] || []
+    return permittedRoles.some((permittedRole) =>
+      user.roles.includes(permittedRole)
+    )
+  })
+}
```
damassi, line 45: @joeyAghion - are you not going to use @dzucconi's nice solution to this TS issue?

src/components/GlobalNav.tsx
joeyAghion: This ☝️ seemed to work.

src/system/index.ts
joeyAghion: Updated with @dzucconi's suggestion, with a tweak or 2.

damassi: Sweet, looks great.

jonallured:
Good to merge @joeyAghion ?

joeyAghion:
Yup thanks @jonallured.

I'd like to follow up with more in-page authorization checks.

PR #100, opened 2022-04-21
feat: use latest version of artsy managed orbs, and track app version in k8s. (PLATFORM-4103)

artsyjian:
feat: use latest version of artsy managed orbs, and track app version in k8s. (PLATFORM-4103)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   cypress: cypress-io/cypress@1
-  hokusai: artsy/hokusai@0.7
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -3,8 +3,8 @@ version: 2.1
 orbs:
   cypress: cypress-io/cypress@1
   hokusai: artsy/hokusai@0.7
```
ovasdi, line 5:
```suggestion
  hokusai: artsy/hokusai@volatile
```
@artsyjian

ovasdi (approved):
🦖

PR #101, opened 2022-04-21
feat: add datadog tracing

mc-jones:
This PR adds datadog tracing to `forque`. 

### Rational
As part of the production-ready checklist, we want to add performance monitoring to `forque`. 

### Solution
`next.js` doesn't support integrations that must be run before any other module (unless you create a custom server) and datadog has a requirement to run before "[any other instrumented module](https://docs.datadoghq.com/tracing/setup_overview/setup/nodejs/?tab=containers#adding-the-tracer-in-code)". 

This is discussed in this [open issue](https://github.com/vercel/next.js/discussions/16600) on the repo, which also includes a suggested workaround, that points to using `node`'s `--require` flag to preload the datadog code before running the app. This seems the most reasonable and low touch approach, especially considering `next.js` even suggests [avoiding using a custom server](https://nextjs.org/docs/advanced-features/custom-server) in their docs. 

I tested this locally by running a local datadog agent in docker and was able to see that it was tracing events from local `forque` in the container's console.

### Consideration
While I tested locally, the setup doesn't exactly match our staging and prod environments, so there is a chance I am missing something here. 

### Post Deploy
- Add `DD_APM_ENABLED` to development `.env` files and staging and prod config maps. 

[PLATFORM-4110]

[PLATFORM-4110]: https://artsyproduct.atlassian.net/browse/PLATFORM-4110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


anandaroop (approved):
Seems like a reasonable solution 👍🏽

PR #106, opened 2022-04-22
chore: improve datadog tracing documentation

mc-jones:
This PR :

1. changes `DD...` to `DATADOG..` to be more consistent with the hokusai config and other datadog related variables.
2. Explicitly sets the `DATADOG_APM_ENABLED` to false, to disable tracing for local development, rather than relying on `undefined`.
2. Adds a comment explicitly stating how the `datadog.js` module is loaded, as this would not otherwise be immediately obvious. 

I'll update the config map with the new key before merging.

### Related PRs
- #101 
- #103


ovasdi (approved):
🦖

PR #114, opened 2022-04-27
chore: add sentry

mc-jones:
This PR adds sentry exception monitoring to `Forque`. For the configuration, I simply followed the `sentry` [docs](https://docs.sentry.io/platforms/javascript/guides/nextjs/) for `next.js`. This handles most of the setup with a few minor required tweaks.

In the interest of keeping this PR small and focused, I have intentionally not added a lot of custom config at this point. This PR should be followed up with separate PRs to customize, as necessary. I have intentionally left the comments generated from the `npx` [script](https://docs.sentry.io/platforms/javascript/guides/nextjs/#configure), which handles the creation of most of these files, as they provide good documentation at this stage. I imagine we will likely want to remove them down the road.

To test the config I set up a project in sentry and threw errors from my local dev instance. They can be seen [here](https://sentry.io/organizations/artsynet/projects/forque-local-test/?project=6361921).

### Considerations
While no further work will need to be done to capture client side errors (not counting custom config), any new API handlers will need to be wrapped in the`withSentry` function to report errors. 

### Follow up 
- [x] Create forque-staging project in sentry
- [x] Add the SDK key to staging env
- [x] Create forque-production project in sentry
- [x] Add the SDK key to production env
- [ ] Delete local testing project


src/pages/api/uploads/[key].page.ts
```diff
@@ -30,4 +31,4 @@ const handler = async (req: NextApiRequest, res: NextApiResponse) => {
   )
 }
 
-export default handler
+export default withSentry(handler)
```
Comment (anandaroop) [inline src/pages/api/uploads/[key].page.ts:34]: So this will be required for every api route we add in the future (but not for ordinary page routes), correct?

anandaroop (approved):
👍🏽

src/pages/api/uploads/[key].page.ts
Comment (mc-jones) [reply src/pages/api/uploads/[key].page.ts:34]: Yup! Thats my understanding. Would be ideal if there was a middleware global error handler that we could hook into (as in express), but for now, it looks like the only way to do it is to add to each route. Something for us to keep an eye out for in future reviews. Saw that there is a [beta for middleware](https://nextjs.org/docs/advanced-features/middleware) which is just generally interesting, but could maybe also allow support for this in the future.

joeyAghion (approved):
What's the default handling of "client" errors (`4xx` status codes or their client-side equivalent) like 404s or forbidden?

I ask because many of our js apps suffer from handling both client- and server-type errors the same way. Ideally, we only record the problems with our code as errors rather than problems with user actions.

(This looks mergeable either way.)

src/pages/api/uploads/[key].page.ts
Comment (damassi) [reply src/pages/api/uploads/[key].page.ts:34]: @mc-jones - we're on the latest next, so this new middleware functionality should work in our app! Not sure if that would simplify this. 

(This current solution isn't _that_ bad, just slightly irksome as something we'll have to remember.)

damassi (approved):
Looks great 👍 Nice and straightforward.

mc-jones:
> What's the default handling of "client" errors (`4xx` status codes or their client-side equivalent) like 404s or forbidden?
> 
> I ask because many of our js apps suffer from handling both client- and server-type errors the same way. Ideally, we only record the problems with our code as errors rather than problems with user actions.

The default should be to report ALL errors and just send them to sentry, which we will need to calibrate. What we can do for the above is to set a filter, in your example on the client config to [filter errors](https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/filtering/) with code.

PR #127, opened 2022-05-03
fix: enable sentry webpack

mc-jones:
This reenabled sentry's webpack plugin after disabling it in #119 due to failing builds. After having created a new integration for each project (required to get an auth token - see [docs](https://docs.sentry.io/product/integrations/integration-platform/#auth-tokens-1)), we can pass the key in directly into the `sentryWebpackPluginOptions`.


joeyAghion (approved):
Does this need to be configured in Circle (just for release-tracking) or also on the deployed applications?

PR #136, opened 2022-05-06
Revert "fix(uploader): swr call uses array as key to impose no cache"

ovasdi:
Revert artsy/forque#133


damassi:
@ovasdi - Mind updating the description with the reason why this needed to be reverted? That way later we'll know not to do it again!

PR #139, opened 2022-05-10
feat: shortcuts

mc-jones:
This PR moves the artsy shortcut widget in `vanity` into `forque`. This is effectively a 1-to-1 transfer, with no significant changes made to the functionality, with one exception: we have decided not to move over the view all page as the value to effort didn't seem to warrant it.

### Considerations
This PR brought up a few pattern question, that I thought would be good to bring to the group in #internal-tools. For example, how much do we want to rely on `swr` vs native `fetch`. How much effort do we want to put in generalizing components for generalized use (e.g., clipboard copy), expectations of testing. 

### Demo
https://www.loom.com/share/38fa82f28e2b4b29b22825c0261d96b5

### TODO:

- [ ]  testing
- [x]  add url validation (if user doesn't include `https://`, the redirect breaks)
- [x]  fix yup validation for required UTMs (campaign, source, medium)
- [x]  Add  clipboard copy functionality

[PLATFORM-4090]

[PLATFORM-4090]: https://artsyproduct.atlassian.net/browse/PLATFORM-4090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/shortcuts/components/CreateOrEditShortcutForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+  const [shortcutResponse, setShortcutResponse] = useState<ShortcutResponse>()
+
+  const handleSubmit = async (values: FormValues) => {
+    const {
+      long,
+      short,
+      source,
+      medium,
+      campaign: campaign,
```
mc-jones, line 58: Will fix.

src/pages/shortcuts/components/CreateOrEditShortcutForm.tsx
```diff
@@ -0,0 +1,246 @@
...
+    medium: "",
+    campaign: "",
+    content: "",
+    term: "",
+  }
+
+  return (
+    <>
+      <Formik
+        initialValues={{
```
damassi, line 128: initialValues={initialValues}

src/pages/shortcuts/index.page.tsx
```diff
@@ -0,0 +1,21 @@
...
+        </Tab>
+        <Tab name="Edit">
+          <EditShortcut />
+        </Tab>
+      </Tabs>
+    </>
+  )
+}
+
+export default ShortcutsPage
```
anandaroop, line 21: This page may be fine as-is... but just FYI one of the quirks of Forque is that page components like this will often (usually?) need to do a little boilerplate to ensure that runtime config vars can be accessed correctly.

```
export const getServerSideProps = () => ({ props: {} })
```

Noted [here](https://github.com/artsy/forque/pull/22), and you'll see this on [several other pages](https://github.com/artsy/forque/search?q=%22export+const+getServerSideProps%22).

mc-jones: Ah, nice call out. Good to know.

src/components/GlobalNav.tsx
```diff
@@ -44,6 +44,7 @@ export const GlobalNav: FC<GlobalNavProps> = ({ user }) => {
             {isPermitted(user, [Action.list, Action.create], "uploads") && (
               <Item href="/uploads">Uploads</Item>
             )}
+            <Item href="/shortcuts">Shortcuts</Item>
```
joeyAghion, line 47: I'd add a role-check here. The API currently [demands](https://github.com/artsy/gravity/blob/e53a61876f7f6aec7146f18355ed62359268a849/app/api/v1/shortcuts_endpoint.rb#L49) the `admin` role.

We've discussed wanting to deprecate that role and _not_ wanting to cargo-cult it into Forque. Up to you if we should take this opportunity to do that or follow-up (by adding a role like `content_manager` perhaps).

joeyAghion (approved):

damassi:
@mc-jones - can you add some screenshots / screencap of this feature? Eager to see what it looks like!

src/pages/shortcuts/components/CreateOrEditShortcutForm.tsx
```diff
@@ -0,0 +1,286 @@
...
+        headers: {
+          "Content-Type": "application/json",
+          "X-Access-Token": accessToken,
+        },
+        method,
+        body: JSON.stringify({
+          short,
+          long: longUrlWithUtm,
+        }),
+      })
```
dzucconi, line 89: As mentioned in Slack: ideally we'd treat Forque similar to Force in that: Metaphysics should be the source of truth for everything. So neither `fetch` nor `swr`, really. We have `shortcuts` in the schema already for fetching. We'll have to add a mutation for creating the shortcut, however.

mc-jones:
> @mc-jones - can you add some screenshots / screencap of this feature? Eager to see what it looks like!

NP. Added loom video to chat.

damassi:
Thanks @mc-jones - loom is pretty sweet. Great work on this! 

After I get back from the dentist lets look at what it will take to move this functionality into MP, but this is a great first pass 👍

mc-jones:
@damassi - Unless there is a major objection, I'm going to merge this PR and we can follow up with the relay work (with more robust testing).

damassi:
Nope, no objection!

PR #158, opened 2022-05-17
feat(auth): raise friendly, meaningful errors when a function is forbidden

joeyAghion:
Here's my attempt to support pages' easily asserting that the authenticated user is authorized. E.g., a user navigating directly to `/shortcuts` should get a meaningful message stating that they don't have the necessary privilege (`content_manager`) rather than encounter mysterious errors when hitting the API.

My [reading](https://nextjs.org/docs/advanced-features/error-handling) indicates that the right way to do this is to embed components in an `ErrorBoundary` like [we already do](https://github.com/artsy/forque/blob/0682e1b57bcec82ecce5e4f50b7f2205eccec920/src/pages/_app.page.tsx#L32). Throwing within an individual page component (like e.g. shortcuts/index) is supposed to be caught and handled at the boundary. The development experience is OK enough:

Screen Shot 2022-05-17 at 4 19 15 PM

But the production experience is unhelpful. The error message is only visible in the browser console, and links to [this documentation](https://nextjs.org/docs/messages/client-side-exception-occurred) indicating that the error somehow wasn't within the error boundary:

Screen Shot 2022-05-17 at 4 37 17 PM

### DIFF
--- a/src/pages/shortcuts/index.page.tsx
+++ b/src/pages/shortcuts/index.page.tsx
@@ -2,8 +2,15 @@ import { Tabs, Tab } from "@artsy/palette"
 import { FC } from "react"
 import { CreateOrEditShortcutForm } from "./components/CreateOrEditShortcutForm"
 import { EditShortcut } from "./components/EditShortcut"
+import { assertPermitted, Action } from "system"
+import type { UserWithAccessToken } from "system"
+import { useSession } from "next-auth/react"
 
 const ShortcutsPage: FC = () => {
+  const session = useSession()
+  const user = session.data?.user as UserWithAccessToken
+  assertPermitted(user, Action.create, "shortcuts")
+
   return (
     <>
       <Tabs>
--- a/src/pages/artists/dedupe/index.page.tsx
+++ b/src/pages/artists/dedupe/index.page.tsx
@@ -1,6 +1,8 @@
 import { useState } from "react"
 import { useMetaphysics } from "hooks"
 import { ArtistList, Skeleton } from "./components/list/ArtistList"
+import { useSession } from "next-auth/react"
+import { Action, assertPermitted, UserWithAccessToken } from "system"
 
 const PER_PAGE = 36
 
@@ -14,6 +16,10 @@ export type RecentArtist = {
 }
 
 export default function Page() {
+  const session = useSession()
+  const user = session.data?.user as UserWithAccessToken
+  assertPermitted(user, Action.list, "artists")
+
   const [page, setPage] = useState<number>(1)
 
   const { data, error, isLoading } = useMetaphysics(
--- a/src/pages/my-collection/index.page.tsx
+++ b/src/pages/my-collection/index.page.tsx
@@ -1,9 +1,15 @@
 import { Text } from "@artsy/palette"
+import { useSession } from "next-auth/react"
 import Head from "next/head"
 import { FC } from "react"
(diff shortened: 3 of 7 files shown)


src/pages/shortcuts/index.page.tsx
```diff
@@ -2,8 +2,15 @@ import { Tabs, Tab } from "@artsy/palette"
...
 import { CreateOrEditShortcutForm } from "./components/CreateOrEditShortcutForm"
 import { EditShortcut } from "./components/EditShortcut"
+import { assertPermitted, Action } from "system"
+import type { UserWithAccessToken } from "system"
+import { useSession } from "next-auth/react"
 
 const ShortcutsPage: FC = () => {
+  const session = useSession()
+  const user = session.data?.user as UserWithAccessToken
+  assertPermitted(user, Action.create, "shortcuts")
```
damassi, line 12: This looks great, though another way to check this would be to use [`middleware`](https://nextjs.org/docs/advanced-features/middleware). That might help surfacing more user-friendly errors to the client.

mc-jones:
I don't think this isn't actually gatting access here, the error we're seeing is just an uncaught exception error, and seems we're not actually hitting the error boundary we've defined (instead we're seeing the [ReactDevOverlay](https://www.npmjs.com/package/@next/react-dev-overlay) error boundary). If you catch the error in the index page and just log it, you can see it is still letting you through to shortcuts. 

```js
try {
  assertPermitted(user, Action.create, "shortcuts")
}  catch(err)  {
  console.log(err)
}
 ```

joeyAghion:
So let me reframe this PR as a question: what is the conventional way to short-circuit a page/route with a deliberate error and message?

I don't think middleware is sufficient, because each page will want to provide different arguments to the assertion.

Does the `throw` need to happen somewhere other than the component (maybe a hook?) to hit the error boundary?

Sorry for the naïve questions but I googled unsuccessfully. Happy to take commits or screen-shares.

damassi:
> I don't think middleware is sufficient, because each page will want to provide different arguments to the assertion.

If you click through and check the linked docs, this (very new) next.js middleware feature can be configured per route. (All good with other implementation tho.)

anandaroop:
I think @mc-jones is right that the `ErrorBoundary` in its current form is not kicking in, thus we are falling through to Next's `ReactDevOverlay`.

I'm pushing up a commit that fixes that `ErrorBoundary` component, so that (with no other changes) we now get this…

Screen Shot 2022-05-19 at 1 21 23 AM

While this is a step in the right direction, it still has a problem: you can't use the nav bar to navigate away from that error page 😬. 

The `ErrorBoundary` doesn't seem to play nice with client-side navigation. There must be a Next-y solution to this, but I don't yet know what it is.

joeyAghion:
Sweet--thanks @anandaroop! The error boundary seems to work more as expected now.

@damassi you're right that middleware could provide this, but since folder structure determine which middleware applies, and different pages within the same folder (e.g., `index.page.tsx` and `[slug].page.tsx`) might want to require different roles, I think I actually prefer asserting in each top-level page. Let me know if you think otherwise.

This last commit adds that to most of the top-level pages. The updated error message looks like:

Screen Shot 2022-05-19 at 3 11 06 PM

Thanks for not laughing at me as I scratch my head about this stack.

damassi:
I like the assert! 🚀

mc-jones (approved):

PR #168, opened 2022-05-20
refactor: generalize isPermitted

mc-jones:
Had some thoughts following #158 but thought it would be better to put a quick draft PR to iterate, rather than delay the previous work from getting into production. 

See comments below for motivation for each change. 

Goal here is just to see if we think this slightly different approach makes sense.


src/components/GlobalNav.tsx
```diff
@@ -38,13 +38,13 @@ export const GlobalNav: FC<GlobalNavProps> = ({ user }) => {
             {isPermitted(user, [Action.dedupe], "artists") && (
               <Item href="/artists/dedupe">Dedupe Artists</Item>
             )} */}
-            {isPermitted(user, [Action.transfer], "my_collection") && (
+            {isPermitted(user, "my_collection") && (
```
mc-jones, line 41: We may have had specific reasoning here, but when I came across this the first time, I was a bit confused as to what to include in each action. So I changed the approach to `isPermitted` to not require an action and we just handle that programmatically within the function itself. Login should be the same: if a user's Role(s) allow them to  perform ANY action within the domain, then show them the nav option.

src/pages/shortcuts/components/EditShortcut.tsx
```diff
@@ -53,6 +53,14 @@ export const EditShortcut = () => {
     }
   }
 
+  if (isPermitted(user, "shortcuts", Action.edit)) {
```
mc-jones, line 56: If you want to gate an individual action, you can pass it in as the third argument and render the component conditionally based on the result.

src/system/index.ts
```diff
@@ -8,69 +8,87 @@ export type UserWithAccessToken = User & {
 // all supported roles
 // NOTE: Do not add Admin to supported roles. It is to be deprectated.
 export enum Role {
-  customer_support = "customer_support",
-  metadata_admin = "metadata_admin",
+  customerSupport = "customer_support",
```
mc-jones, line 11: Since we're in Javascript land we should probably stick to camelCase.

src/system/index.ts
```diff
@@ -8,69 +8,87 @@ export type UserWithAccessToken = User & {
...
+
+  return user.roles.some((role) => {
+    if (action) return actionPermittedByrole(action, role)
+
+    for (const actionEnum in PERMISSIONS[domain]) {
+      if (actionPermittedByrole(actionEnum, role)) return true
+    }
+    return false
   })
 }
```
mc-jones, line 67: Acknowledge this is a bit of a confusion function, but just want to get to work before making it a bit easier to read. 

Also, will need to fix typing issues resulting from not using the Enums properly. Again, just wanted to demo.

src/system/index.ts
```diff
@@ -8,69 +8,87 @@ export type UserWithAccessToken = User & {
...
+    const permittedRoles = buildPermittedRoles(domain)
+
+    const message = `Unauthorized: ${domain} requires role(s): ${permittedRoles.join(
       ", "
     )}. Please contact your product team for assistance.`
     throw new Error(message)
   }
 }
+
+const buildPermittedRoles = (domain: Domain): [string] => {
```
mc-jones, line 84: Since we're no longer passing in the actions as arguments, I need to rebuild the permitted roles to keep the helpful error message. I really liked the specificity of the error message.

src/system/index.ts
```diff
@@ -8,69 +8,87 @@ export type UserWithAccessToken = User & {
...
-  list,
-  dedupe,
-  create,
-  transfer,
-  edit,
+  list = "list",
+  dedupe = "dedup",
+  create = "create",
+  transfer = "tranfer",
+  edit = "edit",
```
mc-jones, line 22: Wondering if we should try to establish a pattern here (limit to crud-like actions, maybe)?  I can see this getting a bit busy over time.

src/system/index.ts
joeyAghion: This is the only thing that I hesitate about, since the specific string values are important (if they're expected to match the token payloads). As long as we _always_ use the string values that way, this is OK.

joeyAghion (approved):

src/system/index.ts
anandaroop: Minor typos here: `dedup` → `dedupe` and `tranfer` → `transfer`

But I take your point about standardizing the actions. This proposal would also rhyme with our [`Ability` class in Gravity](https://github.com/artsy/gravity/blob/main/app/models/domain/ability.rb).

- `dedupe` could reasonably be replaced by `delete`
- not sure about `transfer`, maybe `edit` would suffice?

src/system/index.ts
mc-jones: Fair point. Reverted back to snake case.

mc-jones:
Sorry for how long it took for me to get around to simplifying this implementation. But Now ready for another pair of :eyes.

src/system/index.ts
```diff
@@ -51,37 +53,45 @@ const PERMISSIONS: Record<string, Record<string, Role[]>> = {
   },
   verifications: {
     [Action.create]: [Role.verification_admin],
-    [Action.list]: [Role.verification_admin],
+    [Action.list]: [Role.metadata_admin],
```
joeyAghion, line 56: Intentional change?

src/system/index.ts
mc-jones: Thank @anandaroop! Fixed the typos. Re: standardizing actions, thinks it's worth the discussion, but will move to #internal-tools or a separate PR.

joeyAghion (approved):

PR #176, opened 2022-05-25
fix(env): remove debugging pages

joeyAghion:
These pages were useful when first understanding how to work with `publicRuntimeConfig`, but now that we have [plenty of examples](https://github.com/artsy/forque/search?q=publicRuntimeConfig) I think they can be removed.


anandaroop (approved):
👍🏽

PR #184, opened 2022-06-01
Remove incomplete fields on user edit

jonallured:
In the spirit of impact over perfection I present the user edit page without fields that aren't wired up:

Screen Shot 2022-06-01 at 3 26 00 PM

I think it's worth merging this PR because even in this state the page is usable and allows others to add functionality for working with user records. Later when we can return to this page then we can revert this commit and get a bunch of the paperwork done for free and continue on!


damassi (changes requested):
Why are we deleting this code? It _will_ be used later. We should comment it out. `{/* ... */}` is an easy way to multi-line comments.

jonallured:
@damassi what's wrong with removing it? The code is in git, there's no reason we can't get it back that way, right?

damassi:
It is in git, but the next person that comes along might not remember that and since these fields are pretty straightforward, and any code we might add will resurrect them in a nearly identical format, there's no harm in commenting out. Cut one bit from comment, paste in area outside of comment; repeat.

joeyAghion (approved):

mc-jones (approved):

joeyAghion:
I'm OK with deleting. If an engineer resumed this work _without_ referring to past work... that would be a pretty big failure of institutional knowledge.

I'll merge later unless you modify in the meantime.

damassi (approved):
Alright, 2 over 1 wins 🚪

PR #206, opened 2022-06-10
feat(feature-flags): add new feature-flag page

damassi:
This adds a new Feature Flags section to our Admin app which streamlines the steps required to create flags. 

Related [PR in MP](https://github.com/artsy/metaphysics/pull/4134) needs to be merged first. 

[See this slack thread](https://artsy.slack.com/archives/C0376CFU527/p1655144046876099?thread_ts=1655141259.679009&cid=C0376CFU527) for screencap; internet is currently running too slow!


damassi:
@dzucconi - could you give this a quick review? I knocked it out kinda fast and i'm sure you could add a bit of polish here and there but mostly ready to go.

src/pages/feature-flags/components/CreateFeatureFlag.tsx
```diff
@@ -0,0 +1,249 @@
...
+          })
+
+          setTimeout(() => {
+            sendToast({
+              message: `Successfully created ${values.name}.`,
+              variant: "success",
+            })
+
+            router.push("/feature-flags")
+          }, 1000)
```
dzucconi, line 68: This submits _too_ quickly?

src/pages/feature-flags/components/CreateFeatureFlag.tsx
```diff
@@ -0,0 +1,249 @@
...
+              >
+                Experiment
+              </Checkbox>
+
+              {values.experiment.isEnabled && (
+                <>
+                  {values.experiment.variants.map(
+                    (variant: any, index: number) => {
+                      return (
+                        <Box border="1px solid #ccc" key={index} p={2} my={1}>
```
dzucconi, line 171: `borderColor="black10"`

src/pages/feature-flags/components/CreateFeatureFlag.tsx
```diff
@@ -0,0 +1,249 @@
...
+                </>
+              )}
+            </Box>
+
+            <Separator my={2} />
+
+            <Button
+              loading={
+                // FIXME: Fix this in palette
+                isSubmitting ? ("true" as unknown as boolean) : undefined
```
dzucconi, line 216: What's to fix? https://github.com/artsy/palette/blob/70df27cab4a35c8edb1b04a60996ff5e9be340f6/packages/palette/src/elements/Button/Button.tsx#L31

src/pages/feature-flags/components/CreateFeatureFlag.tsx
```diff
@@ -0,0 +1,249 @@
...
+            </Message>
+          </Form>
+        )
+      }}
+    </Formik>
+  )
+}
+
+const JSONLog: React.FC = ({ children }) => {
+  return <pre>{JSON.stringify(children, null, 2)}</pre>
```
dzucconi, line 246: We really need a `code` font style in Palette IMO

src/pages/feature-flags/components/CreateFeatureFlag.tsx
dzucconi: That's not a real error though. (Like it has absolutely no bearing on anything of consequence)

It can be fixed with 
https://styled-components.com/docs/api#transient-props or https://styled-components.com/docs/api#shouldforwardprop — we shouldn't try to hack around it in the consumers.

src/pages/feature-flags/components/CreateFeatureFlag.tsx
dzucconi: https://github.com/artsy/force/search?q=%22borderColor%3D%22black10%22%22&type=code

damassi: I wonder if prop order matters? 

Update, yep, just checked and this is indeed the issue

src/__generated__/useCreateFeatureFlagMutation.graphql.ts
```diff
@@ -0,0 +1,282 @@
...
+export type FeatureFlagToggleType = "EXPERIMENT" | "RELEASE" | "%future added value";
+export type FeatureFlagVariantWeightType = "VARIABLE" | "%future added value";
+export type AdminCreateFeatureFlagInput = {
+  clientMutationId?: string | null;
+  description?: string | null;
+  impressionData?: boolean | null;
+  name?: string | null;
+  strategy: FeatureFlagStrategyInput;
+  type: FeatureFlagToggleType;
+  variants?: ReadonlyArray<FeatureFlagVariantInputName | null> | null;
```
mc-jones, line 22: We should add the ability to set `stickiness` here, (which actually can _ONLY_ be set via the API) as mentioned in the setup [docs](https://github.com/artsy/force/blob/main/docs/unleash_a-b_testing.md#creating-a-test).

src/pages/feature-flags/components/CreateFeatureFlag.tsx
```diff
@@ -0,0 +1,254 @@
...
+        handleChange,
+        handleBlur,
+        setFieldValue,
+      }) => {
+        // Storing in a ref so that we can log current values to the screen
+        submitData.current = {
+          input: {
+            name: values.name,
+            type: values.experiment.isExperiment ? "EXPERIMENT" : "RELEASE",
+            strategy: {
```
mc-jones, line 90: Are we planning on supporting all the rollout strategies (e.g., `userID`), or just keep limited to a few?

damassi: Now that everything is all setup it should be trivial to add new ones! I just added the ones I know we're currently using, the most basic.

src/__generated__/useCreateFeatureFlagMutation.graphql.ts
damassi: We have a default set over in MP: https://github.com/artsy/metaphysics/blob/main/src/schema/v2/admin/mutations/createFeatureFlagMutation.ts#L92-L95

src/pages/feature-flags/components/CreateFeatureFlag.tsx
mc-jones: Do we want to add the ability to set different strategies for different environments? We are currently just setting the same strategy for both environments, as the default, yeah? Let me know if I'm just missing it.

mc-jones: This also brings up a strategy question around this I'm not totally clear on; while I definitely see this strips out some of the more confusing aspects of the `unleash` native console (e.g., figuring out what a `default` project is, for example), making it simpler to create a feature flag /variant, I'm wondering how many other places we can add value by making out own tool here? The ability to set `stickiness` on a `variant`, removing the step of having to hit the API directly with `curl`, would definitely be a win, but I'm wondering what else we can do here?

damassi: Their REST api is really really nice and very easy to use. Now that we have everything all set up in MP i think we could do anything that we might want with much ease. 

The idea right now however is to not over-engineer anything. Once we start seeing folks mess with other kinds of strategies and stickiness factors and so on, then it might be time to reevaluate or perhaps even turn off in Forque, but for the most part, folks are just adding on/off flags and simple A/B tests! One for dev, one for prod, duplicating effort. Things like different strategies for the same flag per environment is getting a bit specific to develop for right out the gate, but we should def keep an eye on it. It would be a [one line change](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/admin/mutations/createFeatureFlagMutation.ts#L156).

damassi: One thing this did make me think about though is: We should add a note to the create page so that folks know that there's more full functionality over at the main unleash web-app. The idea with Forque was to create something for the other 90%, but that shouldn't limit us should someone have a creative idea around a given test.

PR #208, opened 2022-06-10
feat: create identity verification

mc-jones:
This PR adds the ability to initiate identity verification with just an email address (i.e. it doesn't require the recipient to be an existing artsy user).  It is backed by multiple other PRs associated with the [Jira Epic](https://artsyproduct.atlassian.net/browse/PLATFORM-4162). 

### Demo

https://user-images.githubusercontent.com/38149304/173129271-b7007ae3-7d06-4cdd-81fc-48146a4e0677.mov

### To-Do: 
- [ ] Tests
- [ ] Confirm the limited permission won't block users who should have access to this feater

### Considerations: 

- The functionality is currently scoped to users with `verification_admin` the backend technically also allows the `admin`  role to create identity verifications, but we do not want to bring that role into `Forque` as it is overly broad and needs to be deprecated.

PR #225, opened 2022-06-14
feat: assert permitted for verifications

mc-jones:
This PR follows #208 and enforces `role` permissions when a user navigates directly to `/verifications`.


artsyjian (approved):
lgtm

PR #226, opened 2022-06-15
feat: identity verifications list view

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-4199

Add Verifications List view, per our [design](https://excalidraw.com/#room=f29b1c531adfce1cee60,o1Kx52ea3VotetP6dsal6A).

Differences from the design:
- Does not respect `/verifications/?email=<email>`. Respecting the param presents the [problem of clearing the param](https://artsy.slack.com/archives/CP9P4KR35/p1655302760753719) during tab transitions (between List and Create). As there is not yet a use case for the param, we will re-visit it later on.
- Does not implement buttons to create a new Verification or Override. We will re-visit this later on.

Thanks to @damassi, @dzucconi, @mc-jones, @mzikherman, @MounirDhahri, @joeyAghion  for all the help!

Todo:
---
- [x] Resolve TypeScript warnings
- [x] Add tests


src/pages/verifications/components/VerificationsList.tsx
```diff
@@ -0,0 +1,53 @@
...
+        }}
+        validationSchema={Yup.object().shape({
+          emailInput: Yup.string()
+            .required("An email is required")
+            .email("Please enter a valid email"),
+        })}
+        onSubmit={({ emailInput }) => {
+          setEmail(emailInput)
+        }}
+      >
```
dzucconi, line 28: To me it wouldn't occur to use Formik here since all you're doing is passing an email into some query. I guess the email validation is nice? But just setting `type="email"` on the input will give you some basic native validation.

src/pages/verifications/components/VerificationsDetails.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const VerificationsDetails: React.FC = (props) => {
+  const scanReferences = props.row.original.scanReferences
+  const overrides = props.row.original.overrides
+
+  const StyledBox = styled(Box)`
+    table, th, td {
+      border: 1px solid black;
+    }
+  }
+  `
```
dzucconi, line 15: If we already have a `Table` component; I wouldn't fight its styling. Keep it consistent.

src/pages/verifications/components/VerificationsOverrides.tsx
```diff
@@ -0,0 +1,43 @@
...
+export const VerificationsOverrides: React.FC = (props) => {
+  const overrides = props.overrides
+
+  const onRowClick = () => {
+    // do nothing
+  }
+
+  return (
+    <>
+      Overrides
```
dzucconi, line 13: Presumably this should be wrapped in a `Text` component with some `variant` specified?

src/pages/verifications/components/VerificationsOverrides.tsx
```diff
@@ -0,0 +1,43 @@
...
+          },
+          {
+            Header: "Reason",
+            accessor: "reason",
+          },
+        ]}
+        data={overrides}
+        onRowClick={onRowClick}
+      />
+      <Spacer my={2} />
```
dzucconi, line 40: I'd move this spacer to the outer context. Generally you'd want to avoid external margins of any kind.

src/pages/verifications/components/VerificationsTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+          )}
+        </>
+      )}
+    </>
+  )
+}
+
+const StyledBox = styled(Box)`
+  td {
+    padding-bottom: 1%;
```
dzucconi, line 170: Yeah; we should just have a consistent table style in a single place. Also using % for padding is not recommended.

src/pages/verifications/components/VerificationsDetails.tsx
artsyjian: is there a way to add border via [Table](https://github.com/artsy/forque/blob/main/src/components/Table.tsx)?

with `SyledBox`:
Screen Shot 2022-06-15 at 5 07 57 PM

w/o:
Screen Shot 2022-06-15 at 5 08 36 PM

src/pages/verifications/components/VerificationsList.tsx
artsyjian: oh. using Formik is like a reflex for me now. can consider taking it out. but i want to be consistent with the [Create](https://github.com/artsy/forque/blob/5c09dc5b2ccf0d56864e78c3ed9a2ee349b36392/src/pages/verifications/components/VerificationCreate.tsx#L18) view, so if we do remove Formik, we should remove for both? cc/ @mc-jones

dzucconi: Yeah; I mean if it's just a single input I don't think we need to bring a whole form management library to bear on it. You could use it if you want but I'd get rid of the React state. No need to have two kinds of state.

src/pages/verifications/components/VerificationsOverrides.tsx
artsyjian: good call! done.

src/pages/verifications/components/VerificationsOverrides.tsx
artsyjian: got it. yeah, it actually looks fine w/o this spacer. removed.

src/pages/verifications/components/VerificationsTable.tsx
artsyjian: yeah, down for re-factoring. i find % not easy to work with.

src/pages/verifications/components/VerificationsDetails.tsx
damassi: This is what was throwing that error that we talked about yesterday -- we can't have styled components embedded in other react components. It needs to be top-level, alongside the main component

src/pages/verifications/components/VerificationsList.tsx
damassi: (somewhat insistently): We should stick to Formik here and keep things consistent across all of our forms. This could easily expand. Formik has everything we need, other devs understand it, etc etc.

src/pages/verifications/components/VerificationsTable.tsx
damassi: Would be worth iterating on it in a follow-up PR!

damassi (approved):
This looks truly fantastic @artsyjian 💯 Nice and clean code through and through and everything makes perfect sense. (Also real nice work on your first foray into Relay 👍)

src/pages/verifications/components/VerificationsDetails.tsx
artsyjian: @damassi - ah, i missed this one. shall take it out.

@dzucconi - you mean integrating that styling right in [Table](https://github.com/artsy/forque/blob/main/src/components/Table.tsx)? okay, will do.

dzucconi: Yeah; that way it uses design system tokens and doesn't go off the rails.

src/pages/verifications/components/VerificationsList.tsx
mc-jones: > To put it simply, formik is our form language at Artsy, and even if it is one field, it is idiomatic across the board.

This was my understanding of the pattern we had set, though I think more through osmosis than having seen it said somewhere explicitly. One thing to note with regard to the `VerificationsCreate` component, there is likely to be additional fields added as this is just the first iteration. Nevertheless, I agree there is a debate worth having here about bringing in a full library for inputs that we know will be this simple, though, I'm __slightly__ less concerned about it in an internal admin app. Not a 1-to-1 but thinking of where we have brought in the entirety of `lodash` for one method.

src/pages/verifications/components/VerificationsDetails.tsx
artsyjian: @dzucconi and I paired and re-styled Table. The work is [PR'd](https://github.com/artsy/forque/pull/231) separately. Once that PR is merged, I will update here.

src/pages/verifications/components/VerificationsTable.tsx
artsyjian: Once this [PR](https://github.com/artsy/forque/pull/226/files#r899385337) is merged, I will remove this padding.

src/pages/verifications/components/VerificationsDetails.tsx
artsyjian: done.

src/pages/verifications/components/VerificationsTable.tsx
artsyjian: done.

src/pages/verifications/components/VerificationsTable.tsx
```diff
@@ -0,0 +1,167 @@
...
+                Header: "Scans",
+                accessor: "scanReferences.length",
+              },
+              {
+                Header: "Overrides",
+                accessor: "overrides.length",
+              },
+            ]}
+            data={verifications}
+            renderExpandedRow={(row: any) => {
```
artsyjian, line 135: the `any` is to resolve typescript error:
```
src/pages/verifications/components/VerificationsTable.tsx:139:50 - error TS2339: Property 'scanReferences' does not exist on type '{}'.

139                     scanReferences={row.original.scanReferences}
```

src/pages/verifications/components/VerificationsTable.tsx
```diff
@@ -0,0 +1,167 @@
...
+                <>
+                  <VerificationsDetails
+                    scanReferences={row.original.scanReferences}
+                    overrides={row.original.overrides}
+                  />
+                </>
+              )
+            }}
+            onRowClick={(row) => {
+              //@ts-ignore
```
artsyjian, line 146: can't think of anything better.

src/pages/verifications/components/VerificationsResults.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface VerificationsResultsProps {
+  email: string
+}
+
+export const VerificationsResults: React.FC<VerificationsResultsProps> = (
+  props
+) => {
+  const email = props.email
+
+  const viewerData = useLazyLoadQuery<VerificationsResultsQuery>(
```
artsyjian, line 15: for some reason, the query fires twice on first form submit, producing a "flicker" effect.

src/pages/verifications/components/VerificationsTable.tsx
dzucconi: I'd encourage us to avoid `any` — this wouldn't fly in Force. My inclination is to relax a little for the internal tools but then again I'm wondering why I feel that way.

src/pages/verifications/components/VerificationsTable.tsx
dzucconi: Are the typings accurate for this Table component? It should infer the type.

src/pages/verifications/components/VerificationsTable.tsx
artsyjian: @mc-jones and I [sleuthed on this](https://artsy.slack.com/archives/CA8SANW3W/p1655481328919789). i am down for sleuthing again!

dzucconi: Yeah; there's a huge opportunity for bugs to slip in when using `any` and it's almost never needed. Happy to help with any type errors.

PR #232, opened 2022-06-16
fix(storybook): get Storybook server working again

anandaroop:
https://artsyproduct.atlassian.net/browse/PLATFORM-4240

Slack thread: https://artsy.slack.com/archives/C0376CFU527/p1655401885865179

We have Storybook set up for isolated component development within Forque. 

But the Storybook setup seems to have gotten broken when we enabled absolute project-relative imports in the project, e.g…

```js
import { Foo } from "system"
// as opposed to
import { Foo } from "../../system"
```

This fixes that by touching up the custom Webpack config inside of Storybook's setup. 

As well, it tweaks the existing `GlobalNav` component to work in the Storybook context, as the use of `useRouter` would otherwise cause errors trying to access a null object.

Server output of `yarn storybook`…

|Before|After|
|---|---|
|before|after|

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -21,6 +21,15 @@ module.exports = {
   },
   "webpackFinal": (config) => {
     /**
+     * Allows project-relative absolute imports to work
+     * @see https://github.com/storybookjs/storybook/issues/11639#issuecomment-689835701
+     */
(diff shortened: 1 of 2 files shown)


ovasdi (approved):
👏

PR #241, opened 2022-06-21
feat: enable batch uploads

laurabeth:
PR Type: **Feature**

The need for batch uploads arose during an unrelated ticket, so here we are. 

This PR:
- changes the file input to accept multiple files
- converts the file list to an array
- maps over the array to upload each file individually (as is required by s3)

### DIFF
--- a/src/pages/uploads/components/UploadButton.tsx
+++ b/src/pages/uploads/components/UploadButton.tsx
@@ -4,18 +4,19 @@ import styled from "styled-components"
 import { Uploader } from "./Uploader"
 
 export const UploadButton = () => {
-  const [file, setFile] = useState<File | null>(null)
+  const [files, setFiles] = useState<File[] | null>(null)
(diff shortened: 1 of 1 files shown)


dzucconi (commented):
This looks good and is probably fine to merge but I'd ask you to consider what happens upon success. At the moment we redirect to the file that was uploaded. What happens now? It redirects to the first one that gets uploaded? We should instead of redirecting, load some kind of success state for the uploads with links to each.

laurabeth:
> This looks good and is probably fine to merge but I'd ask you to consider what happens upon success. At the moment we redirect to the file that was uploaded. What happens now? It redirects to the first one that gets uploaded? We should instead of redirecting, load some kind of success state for the uploads with links to each.

Great callout @dzucconi, thank you! Working on this now and will create a follow-up PR shortly.

PR #244, opened 2022-06-21
feat: idv list view to search idvs by user_id param in url query string

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-4095

Extend Verifications List view so that:

- If `user_id` param exists in URL query string, search Identity Verifications by it.
- Present the results and, as before, present the form for search by email.
- If user searches by email, present the new results, but the route (which has `user_id` param) does _not_ change. It seems good to clear `user_id` param when user has searched by email, but we will not concern ourselves with that for now.
- The route also does _not_ change between tab transitions (List<->Create)

No tests for this change since the changes are in top-level components that are still [pending relay testing support](https://github.com/artsy/forque/pull/226#issuecomment-1159014041).

### DIFF
--- a/src/pages/verifications/components/VerificationsList.tsx
+++ b/src/pages/verifications/components/VerificationsList.tsx
@@ -2,10 +2,14 @@ import { Button, Input, Spacer } from "@artsy/palette"
 import { useState } from "react"
 import { Form, Formik } from "formik"
 import * as Yup from "yup"
+import { useRouter } from "next/router"
 import { VerificationsResults } from "./VerificationsResults"
 
 export const VerificationsList: React.FC = () => {
   const [email, setEmail] = useState("")
+
+  const router = useRouter()
+  const [userId, setUserId] = useState(router.query.user_id)
 
   interface InputTypes {
     emailInput: string
@@ -18,6 +22,7 @@ export const VerificationsList: React.FC = () => {
           emailInput: "",
         }}
         onSubmit={({ emailInput }) => {
+          setUserId("")
           setEmail(emailInput)
         }}
         validateOnChange={true}
@@ -47,7 +52,9 @@ export const VerificationsList: React.FC = () => {
           )
         }}
       </Formik>
-      {!!email && <VerificationsResults email={email} />}
+      {(!!email || !!userId) && (
+        <VerificationsResults email={email} userId={userId} />
+      )}
     </>
   )
 }
--- a/src/pages/verifications/components/VerificationsList.tsx
+++ b/src/pages/verifications/components/VerificationsList.tsx
@@ -9,7 +9,7 @@ export const VerificationsList: React.FC = () => {
   const [email, setEmail] = useState("")
 
   const router = useRouter()
-  const [userId, setUserId] = useState(router.query.user_id)
+  const [userId, setUserId] = useState(router.query.user_id as string)
 
   interface InputTypes {
     emailInput: string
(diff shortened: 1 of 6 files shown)


src/pages/verifications/components/VerificationsResults.tsx
```diff
@@ -5,28 +5,34 @@ import { VerificationsTable } from "./VerificationsTable"
...
 export const VerificationsResults: React.FC<VerificationsResultsProps> = (
   props
 ) => {
   const email = props.email
+  const userId = props.userId
 
   const viewerData = useLazyLoadQuery<VerificationsResultsQuery>(
     graphql`
-      query VerificationsResultsQuery($email: String) {
+      query VerificationsResultsQuery($email: String, $userId: String) {
```
artsyjian, line 19: it just works. see [thread](https://artsy.slack.com/archives/CA8SANW3W/p1655840723028429).

src/pages/verifications/components/VerificationsList.tsx
```diff
@@ -2,11 +2,15 @@ import { Button, Input, Spacer } from "@artsy/palette"
...
 import { Form, Formik } from "formik"
 import * as Yup from "yup"
+import { useRouter } from "next/router"
 import { VerificationsResults } from "./VerificationsResults"
 
 export const VerificationsList: React.FC = () => {
   const [email, setEmail] = useState("")
 
+  const router = useRouter()
+  const [userId, setUserId] = useState(router.query.user_id as string)
```
artsyjian, line 12: without the type cast, typescript complains:
```
$ /Users/jxu/code/forque/node_modules/.bin/tsc                                              
src/pages/verifications/components/VerificationsList.tsx:62:40 - error TS2322: Type 'string | string[]' is not assignable to type 'string'.
  Type 'string[]' is not assignable to type 'string'.                                       
                                              
62         <VerificationsResults email="" userId={userId} />  
```
It appears that `router.query.user_id` can return `string | string[]`. Indeed, if the URL has multiple `user_id` params (which is legal it seems), `router.query.user_id` returns an array.

I don't like the casting, b/c in the case of an array, the casting is wrong. What might be a better approach?

dzucconi: If you know that the query param will never be an array: then don't worry about casting it. It's fine. Or just cast it into a string with `${router.query.userId}` — (camel case in query params is fine as well, btw)

artsyjian: I like @mc-jones' approach which covers the edge case of multiple `user_id` params.

But @dzucconi, you are correct, there should be either one `user_id` param or none, never multiple, as we allow an Admin to search IDVs on only one registered user at a time. Okay I will leave the casting the way it is.

The `${}` approach is interesting. It works great, except that when there's no `user_id` param, it returns a string containing `undefined`, which causes Relay to query on a user id of `undefined`. So I don't think I can use it, but it's good to know of this option.

The `user_id` param being snake case is per [ticket description](https://artsyproduct.atlassian.net/browse/PLATFORM-4095). I have no preference at all, so can switch to camel case if there's strong preference.

mc-jones: use of snake case in the ticket was probably just a carry over from `ruby` land. Any redirect to this page will likely come from another `js` client, so probably good to use camel case, as is convention.

mc-jones (approved):

PR #245, opened 2022-06-22
feat (uploads): create separate success behavior for batch upload completion

laurabeth:
The behavior for a single successful file upload is to navigate to view the uploaded file. In the case of a batch upload, we had no defined alternative.

This PR creates the logic to determine whether a single or multiple files were uploaded, and in the case of a batch upload, displays a list of links to the uploaded files. In the case that an upload failed, the display will show only text for the failed item, along with an icon to indicate which items in the batch failed. 

Screen Shot 2022-06-21 at 8 56 58 PM

### DIFF
--- a/src/pages/uploads/components/BatchUploadResult.tsx
+++ b/src/pages/uploads/components/BatchUploadResult.tsx
@@ -1,5 +1,4 @@
-import { Box, CheckCircleIcon, Spacer, Text, XCircleIcon } from "@artsy/palette"
-import Link from "next/link"
+import { Box, CheckCircleIcon, Text, XCircleIcon } from "@artsy/palette"
 import { FC } from "react"
 
 interface BatchResultProps {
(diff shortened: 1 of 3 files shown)


src/pages/uploads/components/BatchUploadResult.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { FC } from "react"
+
+interface BatchResultProps {
+  results: { key: string; status: "success" | "fail" }[]
+}
+
+export const BatchUploadResult: FC<BatchResultProps> = ({ results }) => {
+  return (
+    <>
+      <Spacer mt={4} />
```
dzucconi, line 12: Would move this to the parent.

damassi:
Nice!

jonallured (approved):
LGTM!! 🎉

PR #249, opened 2022-06-23
fix(tests): Upgrades testing libs to fix React v18 related deprecations in tests

anandaroop:
Follow-up to https://github.com/artsy/forque/pull/52, which included an upgrade to React 18

React 18 [changes the way root elements are rendered](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis) and thus requires upgrades to some associated libs, such as our React Testing Library dependencies.

This takes care of these deprecation warnings that are currently displayed during Jest test runs…

Screen Shot 2022-06-23 at 10 14 04 AM


artsyjian (approved):
lgtm

anandaroop:
Note that there are a few unrelated warnings remaining in test runs re: Formik and what-not; this does not address those.

PR #252, opened 2022-06-23
feat: support email query string param for verifications

mc-jones:
This PR enables the `verifications` page to accept an `email` query string param and use it to fill the input. Additionally, if `action=create` is included as a params the component will set the active tab to `VerificationsCreate`

### Considerations
- A follow-up pr will be made in `torque` to redirect admins to `/verifications` when creating a new `IdentityVerification` as an intermediate step until we get the `Users` page completed in `Forque`. 

Relevent slack thread discussion with demo [here](https://artsy.slack.com/archives/CA8SANW3W/p1656007948594589).


ovasdi (approved):

src/pages/verifications/components/VerificationsCreate.tsx
```diff
@@ -3,7 +3,13 @@ import { Form, Formik } from "formik"
 import * as Yup from "yup"
 import { useCreateIdentityVerification } from "../mutations/useCreateIdentityVerification"
 
-export const VerificationsCreate: React.FC = () => {
+interface VerificationsCreateProps {
+  email: string | undefined
```
mzikherman, line 7: #minor rather than `string | undefined` as the type should it be `string` but optional? (so `email?: string`)

mc-jones: Yeah, good call. That's a better approach!

PR #256, opened 2022-06-27
feat(dedupe): enable field-level overrides when merging

anandaroop:
https://artsyproduct.atlassian.net/browse/PLATFORM-4228

This is one of three PRs to enable the [field-level overrides](https://user-images.githubusercontent.com/140521/149867535-309f46de-dfe8-4add-99ca-42f59c2869af.gif) that are a feature of the new deduping UI.

Gravity: https://github.com/artsy/gravity/pull/15477 
Metaphysics: https://github.com/artsy/metaphysics/pull/4189
👉🏽 Forque: https://github.com/artsy/forque/pull/256

This just wires up the `override` subfields from the application state into the payload for the mutation. When it all comes together, looks like this example, where we choose to preserve the  `birthday` from one of the to-be-deleted records —

https://user-images.githubusercontent.com/140521/176051839-fece2b8e-6dbe-4e16-be63-78e725a0d87a.mov

### DIFF
--- a/src/pages/artists/dedupe/components/manager/Confirm.tsx
+++ b/src/pages/artists/dedupe/components/manager/Confirm.tsx
@@ -5,6 +5,7 @@ import { Artist } from "../types"
 import { useMergeArtists } from "../../mutations/useMergeArtists"
 import { useToasts } from "@artsy/palette"
 import { useRouter } from "next/router"
+import _ from "lodash"
 
 interface Props {
   state: State
@@ -43,6 +44,7 @@ export const Confirm: React.FC<Props> = ({ state }) => {
                   input: {
                     goodId: state.goodId!,
                     badIds: state.badIds,
+                    overrides: _.omitBy(state.overrides, _.isEmpty),
                   },
                 },
               })
--- a/src/pages/artists/dedupe/components/manager/Confirm.tsx
+++ b/src/pages/artists/dedupe/components/manager/Confirm.tsx
@@ -5,7 +5,8 @@ import { Artist } from "../types"
 import { useMergeArtists } from "../../mutations/useMergeArtists"
 import { useToasts } from "@artsy/palette"
 import { useRouter } from "next/router"
-import _ from "lodash"
+import omitBy from "lodash/omitBy"
+import isEmpty from "lodash/isEmpty"
 
 interface Props {
   state: State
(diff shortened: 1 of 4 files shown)


src/pages/artists/dedupe/components/manager/Confirm.tsx
```diff
@@ -5,6 +5,7 @@ import { Artist } from "../types"
 import { useMergeArtists } from "../../mutations/useMergeArtists"
 import { useToasts } from "@artsy/palette"
 import { useRouter } from "next/router"
+import _ from "lodash"
```
mc-jones, line 8: Realizing I failed to do this with my [last PR](https://github.com/artsy/forque/blob/main/src/system/index.ts#L2), do we want to aim to set the pattern of importing individually for `lodash` methods to avoid bringing in the whole library to the bundle? So: 

```
import isEmpty from "lodash/isEmpty"
import omitBy from "lodash/omitBy"
```

mc-jones (approved):
Looks great!

src/pages/artists/dedupe/components/manager/Confirm.tsx
anandaroop: Very good q!

I think for that style of import work we'd need to install all the additional discrete packages separately, e.g https://yarnpkg.com/package/lodash.omitby?

Could do that, I guess it's a question of picking your poison in terms of more packages + more upgrades to maintain -vs- bigger bundle size.

For an internal app, the value prop of optimizing for bundle size seemed less obvious to me, but I'm open to persuasion on this one.

mc-jones: I believe support for this was added in v4, so this should work natively for now: https://lodash.com/per-method-packages, in combination with `webpack` tree-shaking. 

I hear you on not wanting to over-optimize for internal apps. The potential value I see outside of the inherent benefits of a smaller bundle is trying to reinforce good patterns. Namely, I don't think it is unusual for engineers to reference what is being done in other newer parts of the codebase to inform how they should write code. It could be good to try and reinforce this pattern to help potentially propagate it across other, maybe even new, repos.

anandaroop: Oh nice! I wasn't aware of this alternative import style. Added in 2f91669

joeyAghion (approved):

joeyAghion:
If this is the final piece of the de-duping puzzle, want to update the nav in this PR as well?

PR #257, opened 2022-06-28
fix: couple of minor improvements to flags UI

mzikherman:
Adds a bit of typing! Also removes re-fetching the whole list after a delete b/c the a navigation is done to refresh the page (w/ fetch policy that refetches the list anyway, so we are double-fetching).

### DIFF
--- a/src/pages/feature-flags/mutations/useDeleteFeatureFlag.ts
+++ b/src/pages/feature-flags/mutations/useDeleteFeatureFlag.ts
@@ -9,9 +9,7 @@ export const useDeleteFeatureFlag = () => {
         $input: AdminDeleteFeatureFlagInput!
       ) {
         adminDeleteFeatureFlag(input: $input) {
-          featureFlags {
-            ...FeatureFlagsTable
-          }
+          success
         }
       }
     `,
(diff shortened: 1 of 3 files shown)


src/pages/feature-flags/mutations/useDeleteFeatureFlag.ts
```diff
@@ -9,9 +9,7 @@ export const useDeleteFeatureFlag = () => {
         $input: AdminDeleteFeatureFlagInput!
       ) {
         adminDeleteFeatureFlag(input: $input) {
-          featureFlags {
-            ...FeatureFlagsTable
```
mzikherman, line 13: Something about our table component is such that, when re-fetching the whole table as the result of a mutation (like in the other flag mutations), the whole table re-renders, which loses state of which rows you've had expanded already.

It feels like there's something missing w/ our table component that lets you properly abstract a row away.

dzucconi (approved):

src/pages/feature-flags/mutations/useDeleteFeatureFlag.ts
damassi: New data = new table render. Expecting sub rows to reconcile new data with old data and track expanded state is asking a bit much, but we should keep an eye on this in the future.

PR #251, opened 2022-06-23
feat(dedupe): commit artist merge mutation via MP

anandaroop:
https://artsyproduct.atlassian.net/browse/PLATFORM-4226

Follow-up to https://github.com/artsy/metaphysics/pull/4156

- Wires up the deduping UI in Forque to that^ new Metaphysics mutation
- Follows existing Forque patterns for mutation handling and toasts

PR #268, opened 2022-07-01
feat: create identity verification overrides

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-4217

This adds support for creation of identity verification overrides to the Verification List view.

### Notable changes to existing components

* `VerificationsDetails` interface was changed to take in fragments as props
* `VerificationsTable_viewer` fragment now spreads Override and ScanReference fragments
* `Table` component uses `overflow: auto` (to prevent text from overflowing the table)

### Other notable mentions

* Following `VerificationsDetails` interface change the `VerificationsList.spec` was removed (along with Fixtures and Types) as it proved too difficult for me to rewrite that test given the new shape. Instead of leaving unused code behind, I decided to remove it entirely.
* Following creation of an _Identity Verification Override_, the top level data _IdentityVerification_ still requires a page refresh for the updated _state_ to show (see demonstration)

### Feature demonstration

https://user-images.githubusercontent.com/29984068/176963290-a2dcef18-3127-40e1-aa6e-e49c823d6198.mov

More details in comments below 👇


src/components/Table.tsx
```diff
@@ -124,6 +124,7 @@ const TableBase = styled(Box)`
     border-bottom: 1px solid ${themeGet("colors.black10")};
     border-left: 1px solid ${themeGet("colors.black10")};
     padding: ${themeGet("space.1")};
+    overflow: auto;
```
ovasdi, line 127: Added this to improve the visual experience. Can scroll sideways to reveal the rest of the content.

Before:
Screen Shot 2022-07-01 at 4 29 10 PM

After:
Screen Shot 2022-07-01 at 4 31 59 PM

src/pages/verifications/components/VerificationsTable.tsx
```diff
@@ -128,22 +109,14 @@ export const VerificationsTable: React.FC<VerificationsTableProps> = (
...
-              },
             ]}
             data={verifications}
             renderExpandedRow={(row: any) => {
               return (
                 <>
                   <VerificationsDetails
-                    scanReferences={row.original.scanReferences}
-                    overrides={row.original.overrides}
+                    identityVerificationScanReferences={row.original}
```
ovasdi, line 118: We are now passing in the whole set of data into the `VerificationsDetails` for both: ScanReferences and Overrides.

src/pages/verifications/components/VerificationsDetails.tsx
```diff
@@ -1,11 +1,13 @@
...
 import { VerificationsOverrides } from "./VerificationsOverrides"
 import { VerificationsScanReferences } from "./VerificationsScanReferences"
-import type { Override, ScanReference } from "./types"
+import { VerificationsOverrides_identityVerification$key } from "__generated__/VerificationsOverrides_identityVerification.graphql"
+import { VerificationsScanReferences_identityVerification$key } from "__generated__/VerificationsScanReferences_identityVerification.graphql"
 
 interface VerificationsDetailsProps {
-  overrides: Override[]
-  scanReferences: ScanReference[]
+  identityVerification:
```
ovasdi, line 8: Originally I made the change to Details which was too confusing so I decided to instead set one prop and use the `|` for types. Previously the _DetailsProps_ was setup this way:

```
interface VerificationsDetailsProps {
  identityVerificationScanReferences: VerificationsScanReferences_identityVerification$key
  identityVerificationOverrides: VerificationsOverrides_identityVerification$key
}

export const VerificationsDetails: React.FC<VerificationsDetailsProps> = (
  props
) => {
  return (
    <>
      <GridColumns>
        <Column span={1}></Column>
        <Column span={10}>
          <VerificationsScanReferences
            identityVerification={props.identityVerificationScanReferences}
          />
          <VerificationsOverrides
            identityVerification={props.identityVerificationOverrides}
          />
        </Column>
      </GridColumns>
    </>
```

However the following seemed very confusing:
`identityVerification={props.identityVerificationScanReferences}`

Does having one prop make it more clear?

src/pages/verifications/components/VerificationsOverridesCreate.tsx
```diff
@@ -0,0 +1,121 @@
...
+                  placeholder="Reason"
+                  name="reason"
+                  type="text"
+                  onChange={handleChange}
+                  value={values.reason}
+                  error={errors.reason}
+                />
+              </Column>
+              <Column span={1}>
+                <Button size="small" type="submit" variant="secondaryOutline">
```
ovasdi, line 111: I really wanted to add a _cancel_ button here (which would remove the form and add _create_ button back) but am not sure how to implement removal of the form. Perhaps an improvement for another PR.

src/pages/verifications/components/VerificationsDetails.tsx
```diff
@@ -16,8 +18,16 @@ export const VerificationsDetails: React.FC<VerificationsDetailsProps> = (
       <GridColumns>
         <Column span={1}></Column>
         <Column span={10}>
-          <VerificationsScanReferences scanReferences={props.scanReferences} />
-          <VerificationsOverrides overrides={props.overrides} />
+          <VerificationsScanReferences
+            identityVerification={
+              props.identityVerification as VerificationsScanReferences_identityVerification$key
```
ovasdi, line 23: Is this ok? Its the same data that gets passed into both `VerificationsScanReferences` and `VerificationsOverrides` from [the Table](https://github.com/artsy/forque/pull/268/files#diff-f7ef92201d6ecf90fa44c0973528cc676745232d2fd5cc3eefbe6655b723278dR117).

src/pages/verifications/components/__fixtures__/verificationOverrideCreateMutationResponse.tsx
```diff
@@ -0,0 +1,30 @@
+export const verificationOverrideSucessResponse = {
```
anandaroop, line 1: #trivial this should prob just be a `.ts` and not a `.tsx` file

src/pages/verifications/components/__tests__/VerificationsCreate.spec.tsx
```diff
@@ -72,8 +70,10 @@ it("sends the correct input with the mutation", async () => {
   )
 })
 
-xit("shows a toast message when an identity verification fails to be created", async () => {
-  mockCreateIdentityVerification.mockImplementation(() => failureResponse)
+it("shows a toast message when an identity verification fails to be created", async () => {
+  mockCreateIdentityVerification.mockImplementation(() => {
+    throw new Error("cats in the server room")
```
anandaroop, line 75: 😹

src/pages/verifications/components/__tests__/VerificationsOverridesCreate.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+            reason: "its a cat",
+            state: "passed",
+          },
+        },
+      })
+    )
+  )
+})
+
+it("shows a toast message when an identity verification fails to be created", async () => {
```
anandaroop, line 80: Oh nice you got the error assertion working for toasts, making a note to study this [for later](https://github.com/artsy/forque/pull/251#discussion_r905169628)…

src/pages/verifications/components/VerificationsTable.tsx
```diff
@@ -128,23 +109,12 @@ export const VerificationsTable: React.FC<VerificationsTableProps> = (
                 Header: "Email",
                 accessor: "email",
               },
-              {
```
ovasdi, line 131: Forgot to mention this.

As a result of spreading the [ScanReferences and Overrides](https://github.com/artsy/forque/pull/268/files#diff-f7ef92201d6ecf90fa44c0973528cc676745232d2fd5cc3eefbe6655b723278dR52) the individual size of those associations can no longer be calculated the same way. Will need to think about how to get this data. Seemed like a nice to have but not essential.

src/pages/verifications/components/VerificationsOverridesCreate.tsx
mc-jones: Seems reasonable to do this in a follow-up!

src/pages/verifications/components/__tests__/VerificationsOverridesCreate.spec.tsx
mc-jones: Wooo. Nice. Same [here](https://github.com/artsy/forque/blob/main/src/pages/verifications/components/__tests__/VerificationsCreate.spec.tsx#L75)

src/pages/verifications/components/VerificationsDetails.tsx
mc-jones: Personally, think the first approach was more clear, as it is more explicit. But I think this seems a matter of preference - others may have stronger opinions.

mc-jones (approved):

src/pages/verifications/components/VerificationsDetails.tsx
ovasdi: Thanks @mc-jones.
I switched back to using explicit props for `VerificationsDetailsProps`. Looking at this again, I think it is more clear having both. See [3261811](https://github.com/artsy/forque/pull/268/commits/32618113f1aceb5cb5b62a4d090d639cdfa3bc40)

src/pages/verifications/components/VerificationsDetails.tsx
ovasdi: That was done to resolve '_is not assignable to type_' and '_Property 'x' is missing in type_' problems. Type assertion is no longer necessary after changes in [3261811](https://github.com/artsy/forque/pull/268/commits/32618113f1aceb5cb5b62a4d090d639cdfa3bc40)

PR #285, opened 2022-07-12
feat: identity verification scan reference ID cell links to netverify

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-4321

During a meeting with T&S it came up that being able to navigate to a verification result from a scan reference is very useful. This PR turns the relevant cell value into a link. Vars already set.


src/pages/verifications/components/VerificationsScanReferences.tsx
```diff
@@ -50,6 +59,15 @@ export const VerificationsScanReferences: React.FC<
           {
             Header: "Jumio ID",
             accessor: "jumioID",
+            Cell: (props: any) => (
```
ovasdi, line 62: Using [column option Cell](https://react-table-v7.tanstack.com/docs/api/useTable#column-options) to form the URL.

joeyAghion (approved):
> Vars already set.

👍

mc-jones (approved):
💯

PR #292, opened 2022-07-14
chore(table): Replace react-table lib with custom table

damassi:
This replaces `react-table` with our own table component (which matches the same api, pretty much). Now it's simpler and has better type-coverage.


mzikherman:
Sweet! Works great locally.

src/pages/verifications/components/VerificationsResults.tsx
```diff
@@ -14,25 +15,34 @@ export const VerificationsResults: React.FC<VerificationsResultsProps> = (
   const email = props.email
   const userId = props.userId
 
-  const viewerData = useLazyLoadQuery<VerificationsResultsQuery>(
-    graphql`
-      query VerificationsResultsQuery($email: String, $userId: String) {
-        viewer {
-          ...VerificationsTable_viewer
+  return (
+    <LazyQueryRenderer<VerificationsResultsQuery>
```
damassi, line 19: Note that this fixed one of [these](https://github.com/artsy/forque/issues/336) suspense boundary issues. 

We should look for other places we use `useLazyLoadQuery` and replace and we should be good.

PR #294, opened 2022-07-18
Identity Verification (index page) tab state, create inputs reset

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-4342

This introduces two (improvements) features into identity verification components:

1. an `onChange` handler introduced to preserve tab state on the index page
2. clear input fields upon successful creation of identity verification

https://user-images.githubusercontent.com/29984068/179601457-bb928814-a345-473d-bb72-c1276c8a3d6f.mov

### DIFF
--- a/src/pages/verifications/index.page.tsx
+++ b/src/pages/verifications/index.page.tsx
@@ -1,5 +1,5 @@
 import Head from "next/head"
-import { Tab, Tabs } from "@artsy/palette"
+import { Tab, TabInfo, Tabs } from "@artsy/palette"
 import { VerificationsCreate } from "./components/VerificationsCreate"
 import { VerificationsList } from "./components/VerificationsList"
 import { useSession } from "next-auth/react"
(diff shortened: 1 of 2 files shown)


src/pages/verifications/index.page.tsx
```diff
@@ -18,15 +19,30 @@ const VerificationsPage: React.FC = () => {
...
         <title>Identity Verifications | Artsy</title>
       </Head>
 
-      <Tabs initialTabIndex={initialTabIndex}>
+      <Tabs
+        initialTabIndex={tabIndex}
+        onChange={(activeTab) => {
+          if ((activeTab as TabInfo).name === "Create") {
+            router.push("/verifications?action=create")
+            router.query.action = "create"
```
anandaroop, line 37: I'm not sure if the `push` and the `action=` are _both_ necessary. Does it work with just the `push`? (I usually see `router.query` as a readonly object, parsed from the current url)

src/pages/verifications/index.page.tsx
```diff
@@ -18,15 +19,30 @@ const VerificationsPage: React.FC = () => {
...
+            setTabIndex(1)
+          } else {
+            router.push("/verifications")
+            router.query = {}
+            setTabIndex(0)
```
anandaroop, line 42: Also curious why this was necessary. Was this state being lost somehow, even after the `push` hack?

src/pages/verifications/index.page.tsx
ovasdi: @anandaroop removed the unnecessary lines. I got slightly confused when testing this change and didn't realize that the state was updated with just `router.push`. Removed the unnecessary lines here https://github.com/artsy/forque/pull/294/commits/0862686372718fa1402cae1cd6e430f6f28a9bce

src/pages/verifications/index.page.tsx
ovasdi: Same reason as https://github.com/artsy/forque/pull/294#discussion_r924803540

damassi (approved):

anandaroop (commented):
👍🏽

PR #302, opened 2022-07-21
Feat: display IDV URL after new IDV is created

pam-:
### Jira
https://artsyproduct.atlassian.net/browse/PLATFORM-4283

### Notable changes
* getting the IDV url from Metaphysics after a successful mutation
* adding a [Palette banner](https://palette.artsy.net/elements/banner/) to display URL

Screen Shot of a banner showing a URL sent to a user for identification verification

### Launch Plan
Will merge the [MP PR first](https://github.com/artsy/metaphysics/pull/4243), then this one upon approval.


src/pages/verifications/components/VerificationsCreate.tsx
```diff
@@ -43,6 +60,18 @@ export const VerificationsCreate: React.FC<VerificationsCreateProps> = (
                   ?.mutationError
               },
             })
+
+            const verificationURL: string =
+              mutationResponse.sendIdentityVerificationEmail
+                ?.confirmationOrError?.identityVerificationEmail
+                ?.verificationURL ?? ""
+
+            verificationURLMessage = {
```
damassi, line 69: Rather than using raw variable assignment we should use React's `useState` hook and set it that way. Something like 
```tsx
const [verificationMessage, setVerificationMessage] = useState(null)

... 

setVerificationMessage({ ... })
```

src/pages/verifications/components/VerificationsCreate.tsx
```diff
@@ -22,16 +29,26 @@ export const VerificationsCreate: React.FC<VerificationsCreateProps> = (
...
+      {verificationURLMessage && (
+        <Banner variant="defaultLight">
+          <p>
+            {verificationURLMessage.message} <br></br>
+            <a href={verificationURLMessage.url}>
+              {verificationURLMessage.url}
+            </a>
+          </p>
+        </Banner>
+      )}
```
damassi, line 41: We should add some margin between this and the form, so that the banner doesn't fit so closely when it appears. 

Our design sys Palette [makes this super easy](https://palette.artsy.net/atoms/spacing/); so just add a `mb={1}` prop to the Banner component.

src/pages/verifications/components/VerificationsCreate.tsx
```diff
@@ -43,6 +60,18 @@ export const VerificationsCreate: React.FC<VerificationsCreateProps> = (
...
               },
             })
+
+            const verificationURL: string =
+              mutationResponse.sendIdentityVerificationEmail
+                ?.confirmationOrError?.identityVerificationEmail
+                ?.verificationURL ?? ""
+
+            verificationURLMessage = {
+              message: "The following verification URL was sent to the user:",
```
ovasdi, line 70: **Suggestion**: I would remove 'user' because not all persons getting verified are Artsy users.

**Thought**: Looking at the message in the image one thing that stood out was that I dont know who the link was sent to. I wonder if displaying the _email_ instead of 'user' would make that message clearer. Seeing as the form is cleared after submit.

pam-: @ovasdi yes! i was actually wondering about being more specific in the message -- i'll update 👍

src/pages/verifications/components/VerificationsCreate.tsx
```diff
@@ -15,23 +21,37 @@ export const VerificationsCreate: React.FC<VerificationsCreateProps> = (
...
     <>
+      {verificationMessage.url && (
+        <>
+          <Banner variant="defaultLight">
+            <p>
+              {verificationMessage.recipient} <br></br>
+              <a href={verificationMessage.url}>{verificationMessage.url}</a>
+            </p>
+          </Banner>
+          <Spacer my={2} />
```
damassi, line 42: Nice, you read my mind 😄 I was just about to mention this component and now I see it here 👍

damassi (approved):
Nice first contribution to Forque @pam- 👍 Looks great.

ovasdi (approved):
LGTM. Nice job!

The [failure](https://app.circleci.com/pipelines/github/artsy/forque/1127/workflows/fbc57c96-dfbd-4a9d-bb39-fac3cc0f916f/jobs/3124?invite=true#step-105-7) is a result of schema change in MP. You can either check the updated schema into this PR or wait for the automation to sync the schema (after MP PR is merged) for CI to be 🟢 .

PR #334, opened 2022-08-16
feat: add collector profile to user view [PLATFORM-4426]

mdole:
This PR builds on https://github.com/artsy/metaphysics/pull/4308 and is a further branch on https://github.com/artsy/forque/pull/338. It adds a new `Collector Profile` tab to Forque, which includes its own form for sending mutations through Metaphysics.

Screen Shot 2022-08-19 at 16 20 48

Jira ticket: https://artsyproduct.atlassian.net/browse/PLATFORM-4426


src/pages/users/[userId].page.tsx
```diff
@@ -42,6 +45,15 @@ const User: React.FC<UserProps> = ({ user }) => {
             postalCode: user.saleProfile?.zip || undefined,
             country: user.saleProfile?.country || undefined,
           },
+          collectorProfile: {
+            companyName: user.collectorProfile?.companyName || undefined,
```
Comment (damassi) [inline src/pages/users/[userId].page.tsx:49]: Are all of these `|| undefined`'s necessary? that seems a bit like a code-smell, since if `collectorProfile` is `null`, then companyName will also be `null`. 

Also, better to use `null` rather than `undefined`, just in general. Null is value of "nothing" (which we're passing into the field), where `undefined` is a complete lack of a value (it's undefined, no definition, is abstract)

src/pages/users/components/form/index.tsx
```diff
@@ -8,7 +9,10 @@ export const UserForm: React.FC = () => {
       <GridColumns>
         <Column span={4}>
           <UserInfoForm />
+          <Spacer my={4} />
           <UserSaleProfileForm />
+          <Spacer my={4} />
+          <CollectorProfileForm />
```
damassi, line 15: Not suggesting we update this here, but as a tip to make the code a bit more clean can always use the [`<Join>` component](https://palette.artsy.net/elements/join/):

```tsx
<Join spacer={<Spacer my={4} />}>
  <UserInfoForm />
  <UserSaleProfileForm />
  <CollectorProfileForm />
</Join>
```

That way you can keep adding sections without also having to keep adding spacers.

mdole: ah yeah thanks for the reminder! we ended up scrapping this layout and deleting this file, but will use that elsewhere if applicable :)

src/pages/users/[userId].page.tsx
Comment (mdole) [reply src/pages/users/[userId].page.tsx]: yeah I'm definitely guilty of cargo-culting these 😅 I'll see if I can improve them

Comment (mdole) [reply src/pages/users/[userId].page.tsx]: Improved in [e35be8a](https://github.com/artsy/forque/pull/334/commits/e35be8a0b449eb97849ac3ba08121b0e2516cf73)

mdole:
Alright, I think this is in reasonable shape (if we don't talk about specs, which don't exist for this part of the app yet 😅).

One thing I noticed about tabs + forms: if you send a mutation, switch tabs, and switch, back, you'll still see the "old" content in the form. If you refresh the page you'll load in fresh values. Worth considering if we should fix this or perhaps just put a warning label saying "update complete, note that if you switch tabs and switch back it will look like nothing happened"

profile-update

For next steps: when @mc-jones gets back next week, I'll leave it up to him if he wants to just merge this PR (since it has all of the changes from https://github.com/artsy/forque/pull/338) or make further changes to that one, merge it, and then rebase this one. Could go either way!

mc-jones:
Just curious if it was intentional to bring in the dependabot bumps to `terser` ([0d15c78](https://github.com/artsy/forque/pull/334/commits/0d15c78dda4fea86e01297f0737a0dd2c8ff3d65)) and `next-auth` ([11b0c2f](https://github.com/artsy/forque/pull/334/commits/11b0c2f1799784be4e310ebe32637b64ddce7dfa)) into this PR?

mc-jones:
Hm. Interesting. Looks like it's including commits already in `main`. Will squash as there are no conflicts so should be safe.

PR #335, opened 2022-08-16
fix: use key attribute to reset ErrorBoundary

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-4448

This resolves the issue where a user is not able to navigate away from an ErrorBoundry rendered page via navigation links. 

The cause of the issue seems to be that the ErrorBoundary is reused (across renders) however the _state_ (`hasError`) has no way of changing once set. To address this issue (following [this advice](https://github.com/bvaughn/react-error-boundary/issues/23#issuecomment-425470511)) I chose to leverage the _Key_ attribute passing in the _location_, which resets the ErrorBoundary when navigating away from the error page.

https://user-images.githubusercontent.com/29984068/184986141-2d9c7c6c-1b2d-43d4-a27b-d1384857fc57.mov


ovasdi:
After making this change, I noticed the following error in dev mode: "_Uncaught Error: The server could not finish this Suspense boundary, likely due to an error during server rendering. Switched to client rendering._"

Looks like a known and still [open issue](https://github.com/vercel/next.js/issues/35564) that affects nextjs `v12.1.4` and prior.

damassi:
Can I just take a moment and say that I viscerally hate react suspense? The react team has engineered a complexity monster and they should have broken this behavior off into an adjacent npm package. I just ran into this issue too, unrelated to this PR, which I could not figure out how to fix.

mc-jones (approved):
👍 . Thanks for the quick resolution on this @ovasdi.

mc-jones:
> Looks like a known and still https://github.com/vercel/next.js/issues/35564 that affects nextjs v12.1.4 and prior.

We should aim to track issues like this while they come up and before the context is lost and before they proliferate. If the fix is in `12.5` (took a quick look at the [changelog](https://github.com/vercel/next.js/releases/tag/v12.2.5), but wasn't clear to me) we should ticket it out, or if it is still being implemented maybe add a GitHub issue on the repo?

mc-jones:
Minor, but should the PR `type` be `fix` here?

ovasdi:
> We should aim to track issues like this while they come up and before the context is lost and before they proliferate. If the fix is in 12.5 (took a quick look at the [changelog](https://github.com/vercel/next.js/releases/tag/v12.2.5), but wasn't clear to me) we should ticket it out, or if it is still being implemented maybe add a GitHub issue on the repo?

@mc-jones I upgraded (locally) to the _latest_ nextjs release to see if the issue is still present and even with the suggested (temp) workaround of adding `experimental.runtime: 'nodejs'` to the config, I am still seeing that error. Opened [an issue](https://github.com/artsy/forque/issues/336) to track this.

PR #352, opened 2022-08-23
feat: Migrate Admin View to Forque

mc-jones:
This PR migrates over the following fields from the User's App Admin section in `Torque`:
- [x] Two-Factor Authentication Enabled
- [x] Last Sign-in
- [x] Number of sign-ins
- [x] Name
- [x] Email
- [x] Email Confirmation
- [x] Phone Number
- [x] Roles
- [x] Admin Notes
- [x] Partner Access

### Related PRs
- https://github.com/artsy/metaphysics/pull/4304

### Follow-up Items:
- Better error messages. This will require some modifications in Metaphysics to ensure all queries are responding with properly formatted errors. 
- We will need to follow up with tests as the user App is further developed.

[PLATFORM-4430]

[PLATFORM-4430]: https://artsyproduct.atlassian.net/browse/PLATFORM-4430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/users/components/form/UserAccountForm.tsx
```diff
@@ -45,42 +65,86 @@ export const UserAccountForm: React.FC<UserProps> = ({ user }) => {
...
+        phoneNumber: user.phone ?? "",
+        roles: user.roles ?? [],
+        secondFactorEnabled: user.secondFactorEnabled ?? false,
+        newAdminNote: "",
       }}
-      validationSchema={userValidationSchema}
+      validationSchema={Yup.object({
+        email: Yup.string().email(),
+      })}
       onSubmit={async (values) => {
```
mc-jones, line 84: @damassi - I'm planning on "drying up" this `onSubmit` function but before I do, wanted to check if I am simply missing a simpler pattern to handle multiple mutations firing from one `Formik` submission and still be able to differentiate individual errors (this is why I moved away from the previous `Promise.all` implementation).

src/pages/users/useUserFormContext.tsx
```diff
@@ -48,10 +47,9 @@ export interface UserFormValues {
   notes?: string
 
   // Admin
-  emailConfirmedAt?: string
   location?: string
   secondFactorEnabled?: boolean
-  roles: (string | null)[]
+  roles?: any
```
mc-jones, line 52: Just noting this file will be deleted so Any won't stay and it will have prop typing.

src/pages/users/components/form/UserAccountForm.tsx
damassi: Nope! This looks great (and I would suggest DRYing it up further; it's all very literal and clear as it is). One thing I would have done differently would have been to use multiple `try/catch` vs vanilla promises but that's all very surface-level and the vanilla promises are readable, and i think this form is also a nice way to achieve parallization without `Promise.all`.

damassi: (Seeing vanilla promises used here in this way makes me want to return to some recent auction code I've written where there were a series of `await`ed try/catch blocks (needed to distinguish between various errors, as above) and rewrite with vanilla promises to speed things up)

src/pages/users/helpers/useUpdateUserRoles.tsx
```diff
@@ -0,0 +1,52 @@
...
+                id: userId,
+                role: role!,
+              },
+            },
+          })
+        )
+      })
+    })
+
+    return [...deleteRolePromises, ...addRolePromises]
```
damassi, line 48: Nice 👍

src/pages/users/components/form/UserAccountForm.tsx
mc-jones: Awesome. Thanks, @damassi! Will leave it as is except for `try/catch` which I just pushed up.

damassi (approved):
👌

src/pages/users/components/form/UserAccountForm.tsx
```diff
@@ -45,42 +65,86 @@ export const UserAccountForm: React.FC<UserProps> = ({ user }) => {
...
+          Promise.all(
+            updateUserRoles(user.internalID, user.roles, values.roles)
+          )
+            .then(() => {
+              sendToast({ variant: "success", message: "User roles updated" })
+            })
+            .catch(() => {
+              sendToast({
+                variant: "error",
+                message: "Failed to update User Roles",
```
ovasdi, line 141:
```suggestion
                message: "Failed to update user roles",
```
**nit** To stay consistent with other error messages where only the first letter is capitalized?

ovasdi (approved):
LGTM. Awesome job!

src/pages/users/components/form/UserAccountForm.tsx
mc-jones: Good call. Just fixed with [7a1935f].(https://github.com/artsy/forque/pull/352/commits/7a1935f091d67df65270f189b8b1401e85726182)

PR #356, opened 2022-08-25
feat: add user sale profile tab

mc-jones:
This PR creates a view for the User's `UserSaleProfile`. 

### Considerations
After [speaking](https://artsy.slack.com/archives/C0376CFU527/p1661355122544729) with Collector Services, it was agreed that in the first iteration of this view, we should just display all fields from the `UserSaleProfile` model, but not implement the ability to edit them;  it is unclear if anyone actually edits these fields from the Admin app. However, I opted to display them in a disabled `Input` component to give future developers the ability to quickly developed and implement a mutation for these fields. This admittedly come with UI tradeoffs, but I think it is acceptable at this point and will be quicker to refactor to basic text than vice versa.

Screen Shot 2022-08-25 at 4 31 51 PM

### Follow Up
- Test coverage for User's page ([ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-4515))


ovasdi (approved):
LGTM

PR #358, opened 2022-08-26
feat: add data transfer opt out checkbox under Security tab

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-4431

This PR adds a Security tab and a Data Transfer Opt Out checkbox that mutates the `dataTransferOptOut` field when clicked.

Screen Shot 2022-08-25 at 10 41 28 PM

Writing a test for this has proven to be quite challenging for me. Is this test valuable and worth adding?


src/pages/users/components/__tests__/Security.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+  },
+}))
+
+jest.mock("../../mutations/useUpdateUser", () => ({
+  useUpdateUser: () => ({ submitMutation: mockUpdateUserMutation }),
+}))
+
+describe("Security tab", () => {
+  it("mutates user when data transfer is checked", async () => {
+    const user: any = {
```
ovasdi, line 22: Security Tab [component](https://github.com/artsy/forque/pull/358/files#diff-f3293a5acf0f5a94dc64c3e437c094c25042df5a8a4043e18733f1496363c9c7R6) takes in `UserProps` and I haven't figured out to initialize this prop without using `any`.

Should I add a `types.ts` and define a `User` type? Then create fixture similarly how [artistsWithDupes](https://github.com/artsy/forque/blob/main/src/pages/artists/dedupe/components/manager/__fixtures__/artist-with-dupes.ts) is setup?

src/pages/users/components/__tests__/Security.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+      internalID: "id",
+      dataTransferOptOut: false,
+    }
+
+    render(<Security user={user} />)
+
+    expect(screen.getByText("OPT-OUT of Data Transfer")).toBeVisible()
+
+    // eslint-disable-next-line testing-library/no-unnecessary-act
+    await act(async () => {
```
ovasdi, line 32: There was a warning "Warning: An update to Security inside a test was not wrapped in act". So I added `act` as suggested by the console.

ovasdi: Is there another way I should handle this?

damassi: Seems ok to me to follow the warnings! (As a side note, i need to look at tests again in forque, specifically around Relay)

src/pages/users/components/__tests__/Security.spec.tsx
damassi: If the types aren't defined in the actual component in such a way that it can be reused here, i wouldn't sweat using `any` in a test... but that's just me.

damassi (approved):

src/pages/users/components/Security.tsx
```diff
@@ -0,0 +1,55 @@
...
+              await submitUserMutation({
+                variables: {
+                  input: {
+                    id: user.internalID,
+                    dataTransferOptOut: selected,
+                  },
+                },
+              })
+            } catch (error) {
+              console.error("[forque] Error updating user:", error)
```
mc-jones, line 39: Should we add user feedback (via `useToast`) to let them know if the update succeded or failed? Right now I believe the mutation could fail but the UI would reflect a successful update by flipping the checkbox, correct?

src/pages/users/components/Security.tsx
```diff
@@ -0,0 +1,55 @@
...
+            }
+            setIsLoadingDataTransfer(false)
+            setEnabledDataTransfer(!enabledDataTransfer)
```
mc-jones, line 42: Do we want to move this into the `try` block so it doesn't execute on a failure case?

src/pages/users/components/Security.tsx
ovasdi: Yep. Thanks for catching that.

ovasdi: @mc-jones addressed in https://github.com/artsy/forque/pull/358/commits/ac474289226388a169abb589eff6e9882c3463df

src/pages/users/components/Security.tsx
ovasdi: addressed in https://github.com/artsy/forque/pull/358/commits/ac474289226388a169abb589eff6e9882c3463df

src/pages/users/components/Security.tsx
```diff
@@ -0,0 +1,72 @@
...
+              sendToast({
+                variant: "success",
+                message: `${
+                  enabledDataTransfer ? "Disabled" : "Enabled"
+                } Data Transfer`,
+              })
+            } catch (error: any) {
+              console.error("[forque] Error updating user:", error)
+
+              setIsLoadingDataTransfer(false)
```
mc-jones, line 52: I should have been more specific about referring to just the `setEnabledDataTransfer` call when mentioning moving! Minor, but if you wanted you could leave the `setIsLoadingDataTransfer` after the `catch` block and only have to call it once. An error in a `try/catch` doesn't exit the outer function, so anything after the `catch` block will still execute. So code that should execute regardless of how the mutation resolves can still be in the function scope - in this case, the `onClick` callback.

mc-jones:
Nice job 👍

mc-jones (approved):

PR #361, opened 2022-08-30
feat(autocomplete): Add relay example + LazyQueryRenderer

damassi:
This builds on the work that @anandaroop did over in [this branch](https://github.com/artsy/forque/tree/anandaroop/relay-take-2) and finishes it up by establishing a pattern for working with these kinds of suspense-based components. 

Pattern should solve other similar use-cases where we need to do relay refetches in a more client-side-centric way, but still need to support SSR rendering.

autocomplete

### Problem 

- Next.js is SSR based
- For components that have simple `useLazyLoadQuery` mechanisms, which only fire on changes on the client, we need to account for that
- If we don't do that, then it will lead to SSR hydration mismatch warnings, vaguely suggesting we "wrap state change in useTransition"
- Another thing: `useLazyLoadComponent` needs to be wrapped in a suspense boundary, for when things are loading
- There's a problem however: What happens when we have a component like `<AutoComplete />` from palette, with an independent `loading` mechanism that's controlled programmatically via a prop? 
- If we were to wrap our relay backed `ArtistAutoComplete` component in a suspense boundary, every time we typed it would show the suspense boundary loader, hiding the main component:
```tsx
<Suspense fallback="loading...">
  <ArtistAutoComplete />
</Suspense>
```

### Explanation of Solution

- In order to fix, we need to separate concerns between UI and suspense-based data-fetching, and we had to replace the use of `useLazyLoadQuery` with `useQueryLoader`
- The `useQueryLoader` is effectively lazy, but needs to be called directly (vs `useQueryLoader` which is called on mount, leading to SSR suspense errors)
- In order to tackle programmatic loading states, suspense-based fetching via relay needs to be isolated within its own component, and everything else in another component. Lets call it `<ArtistFetcher />`. It doesn't actually render UI, and returns `null`. It uses a hook to fetch data which it returns via a callback prop and exists only in order to satisfy the strict requirement that _something be wrapped in `<Suspense>`_ 
- Then we wrap the `<ArtistFetcher />` around a suspense boundary, leaving everything else safe to render during the SSR pass
```tsx
<Suspense fallback={null}>
  <ArtistFetcher query={query} onChange={handleChange} />
</Suspense>

<AutoComplete ... />
```
- Now, how does one trigger the loading state in the `AutoComplete` component without an easy react-level API method that taps into suspense's "is suspended" state? (😢)
- We use _yet another_ faux-component switch that toggles loading state, and pass it into the suspense fallback prop. Full example: 
```tsx
const ArtistAutoComplete = () => {
  const [query, setQuery] = useState("")
  const [isLoading, setIsLoading = useState(false)
  
  const LoadingSwitch = () => {
    useEffect(() => {
      setIsLoading(true)
    }, [])
    
    return null 
  }

  const handleChange = (options) => {
    setOptions(options)

    // By the time we've received the data, loading is complete
    setIsLoading(false)
  } 
  
  return (
    <Suspense fallback={<LoadingSwitch />}>
      <ArtistFetcher query={query} onChange={handleChange} 
    </Suspense>

    <AutoComplete onChange={setQuery} ... />
  )
}
```

Pretty straightforward once React suspense is properly understood. 

_but_

WHY 😭 This new react suspense API design is forcing devs to write faux components all over in order to do simple things. It's bad, bad news. 

### QueryRenderer 

Now, to make this all usable and sane, a new utility component that mirrors the legacy Relay API called `<LazyQueryRenderer>`:

```tsx
<LazyQueryRenderer
  query={graphql`
    query MyQuery(id: String!) {
      artist(id: $id) { 
        name
      }
    }
  `
  variables: {
    id: 'andy-warhol'
  }
  render={({ isLoading, data }) => {
    if (isLoading) {
      return <>Loading!</>
    }
    
    
    return (
      <>
        Hello {data.artist.name]
      </>
    )
  }}  
/>
```

cc @artsy/grow-devs


laurabeth:
> react suspense API design is forcing devs to write faux components all over

I haven't thoroughly explored the code in this PR as yet, but this makes me think we should create a set of common placeholders. Is that possible? Will dig in a little further and sneak-delete this comment if I figure out why we can't. 😝

src/system/relay/QueryRenderer/QueryRenderer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  variables,
+  render,
+}: QueryRendererProps<T>) => {
+  const [data, setData] = useState(null)
+  const [isLoading, setIsLoading] = useState(false)
+
+  useEffect(() => {
+    if (data) {
+      setIsLoading(false)
+      setData(data)
```
anandaroop, line 30: If `data` is already set on 28 why do we set it again on L30?

src/system/relay/QueryRenderer/QueryRenderer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  query: GraphQLTaggedNode
+  placeholder?: JSX.Element
+  render: (props: {
+    isLoading: boolean
+    data: T["response"] | null
+  }) => React.ReactNode
+  variables?: T["variables"]
+}
+
+export const QueryRenderer = <T extends OperationType>({
```
anandaroop, line 19: Do you think we should squat on the old `QueryRenderer` name or give this a name that signals it's our own convenience invention? 

(Wondering if this will confuse future folks we see a `QueryRenderer` at the call site)

anandaroop (commented):
Hats off to you for fighting through this! This was incredibly challenging when I last looked at it.

Some of my concerns are:

- How easy will it be for devs to use this kind of component when doing feature work? 
  - ✅ Seems like that's fine, as the invocation is just as simple in the [old](https://github.com/artsy/forque/pull/361/files#diff-372fa2a7ea0a4da7f1a93843bcd61450ddaf7a4ac8099aac03b66b96d1d3c116R36) and [new](https://github.com/artsy/forque/pull/361/files#diff-372fa2a7ea0a4da7f1a93843bcd61450ddaf7a4ac8099aac03b66b96d1d3c116R30) versions.

- How easy will it be for devs to create similar autocomplete components?
  - 😅 This will obv be a less common occurrence, but the amount of somewhat mysterious boilerplate here does worry me. I guess one thing to do is let someone else similarly convert [the existing `UserAutocomplete`](https://github.com/artsy/forque/tree/main/src/components/autocomplete) to spread this knowledge around.

src/system/relay/QueryRenderer/QueryRenderer.tsx
damassi: I thought about this too, and you know how auto-import can get confused... Suggestion for a name @anandaroop?

src/system/relay/QueryRenderer/QueryRenderer.tsx
damassi: Because when the component initializes, `data` is null. In that case we don't want to fire an update. And if a field comes back `null` it would be `data.artist = null` (or whatever) which the consumer can check against up the chain where responsibility is assumed.

damassi:
> This will obv be a less common occurrence, but the amount of somewhat mysterious boilerplate here does worry me

This is understandable. The issue here is two things: React 18 concurrency -- which is a whole new way of thinking about react -- and Next.js SSR. The solution here is the simplest I could come up with based on a few competing options, and that's why I wanted to write out the full reasoning in the PR description. There's no way to get around it; for as long as we're going to use Relay, Suspense has to be understood; and for as long as we're going to use Next, SSR has to be understood. Combine the two together and this is what you get, which in the end is fundamentally quite simple (assuming the two concepts above are understood). 

I _can_ see this pattern / component being used in lots of other places though. We have a ton of components in Palette with programmatic loading states. In each and every case we'll have to use something like this. 

I'd be happy to write out an example gist of the pattern if that would be helpful.

src/system/relay/QueryRenderer/QueryRenderer.tsx
mzikherman: I would definitely vote to not override existing Relay/React names, otherwise it will get confusing to a more casual reader (plus the auto-import). Abstractions on abstractions!

mc-jones: > I would definitely vote to not override existing Relay/React names, otherwise it will get confusing to a more casual reader (plus the auto-import). Abstractions on abstractions!

Second this.

mc-jones:
> I'd be happy to write out an example gist of the pattern if that would be helpful, if the PR description + code isn't sufficient.

I think future contributors would appreciate this. I always like being given the option of seeing quick references to patterns to develop quickly, while being given the information to go deeper into the motivation/decision making (i.e. this PR).

src/system/relay/QueryRenderer/QueryRendererFetcher.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [queryRef, loadQuery] = useQueryLoader(query)
+
+  useEffect(() => {
+    loadQuery(variables)
+  }, [loadQuery, variables])
+
+  return (
+    <>
+      {queryRef && (
+        <Response
```
damassi, line 29: The reason why this is necessary is because some relay hooks aren't composable. Meaning: 
```tsx
const useFetchData = () => {
  const [queryRef, loadQuery] = useLoadQuery(query)
  
  // Error! queryRef is undefined, because loadQuery hasn't been called yet
  const data = usePreloadedQuery(query, queryRef) 
}
```
Since one can't conditionally render hooks, one has to do it another way, like this. 

Not sure why the relay team doesn't allow one to pass a null queryRef into `usePreloadedQuery`, and just return `null` or `undefined`; its extremely common for devs to do something like this in their components 

```
if (!data) {
  return null
}
```

Because of this, it breaks the nice feel of hooks, where one hook should be composable into another and so on. What relay has designed here is a hook that _has_ to be tied to a component. I'm sure they have their reasons, but again, weird API design leading to weird code for commonly expressed things.

damassi: In any case, this approach makes the code a bit simpler elsewhere, since we no longer have to worry about using Next.js's dynamic import function to import the module.

src/components/LazyQueryRenderer.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const [queryRef, loadQuery] = useQueryLoader(query)
+
+  useEffect(() => {
+    loadQuery(variables)
+  }, [loadQuery, variables])
+
+  return (
+    <>
+      {queryRef && (
+        <Response
```
damassi, line 66: The reason why this is necessary is because some relay hooks aren't composable. Meaning: 
```tsx
const useFetchData = () => {
  const [queryRef, loadQuery] = useLoadQuery(query)
  
  // Breaking runtime error! queryRef is undefined, because loadQuery 
  // hasn't been called yet. Once its called and data fetches, it will 
  // populate queryRef which one can then use to actually read the data
  // from the relay store.
  const data = usePreloadedQuery(query, queryRef) 
}
```
Since one can't conditionally render hooks, one has to do it another way, like it's done in this PR. 

Not sure why the relay team doesn't allow one to pass a null queryRef into `usePreloadedQuery`, and just return `null` or `undefined` from the output; its extremely common for devs to do something like this in their components 

```
if (!data) {
  return null
}
```

Because of this, it breaks the nice feel of hooks, where one hook should be composable into another and so on. What relay has designed here is a hook that _has_ to be tied to a component. I'm sure they have their reasons, but again, weird API design leading to weird code for commonly expressed things.

damassi: In any case, this approach makes the code a bit simpler elsewhere, since we no longer have to worry about using Next.js's dynamic import function to import the module.

src/system/relay/QueryRenderer/QueryRenderer.tsx
damassi: Renamed `QueryRenderer` to `LazyQueryRenderer` (feel free to unresolve this convo if this doesn't work!)

src/components/LazyQueryRenderer.tsx
```diff
@@ -0,0 +1,91 @@
...
```
damassi, line 58: The reason why this is necessary is because some relay hooks aren't composable. Meaning: 
```tsx
const useFetchData = () => {
  const [queryRef, loadQuery] = useLoadQuery(query)
  
  // Breaking runtime error! queryRef is undefined, because loadQuery 
  // hasn't been called yet. Once its called and data fetches, it will 
  // populate queryRef which one can then use to actually read the data
  // from the relay store.
  const data = usePreloadedQuery(query, queryRef) 
}
```
Since one can't conditionally render hooks, one has to do it another way, like it's done in this PR. 

Not sure why the relay team doesn't allow one to pass a null queryRef into `usePreloadedQuery`, and just return `null` or `undefined` from the output; its extremely common for devs to do something like this in their components 

```
if (!data) {
  return null
}
```

Because of this, it breaks the nice feel of hooks, where one hook should be composable into another and so on. What relay has designed here is a hook that _has_ to be tied to a component. I'm sure they have their reasons, but again, weird API design leading to weird code for commonly expressed things.

damassi: In any case, this approach makes the code a bit simpler elsewhere, since we no longer have to worry about using Next.js's dynamic import function to import the module.

src/components/LazyQueryRenderer.tsx
```diff
@@ -0,0 +1,90 @@
+import { Suspense, useEffect, useState } from "react"
+import { PreloadedQuery, usePreloadedQuery, useQueryLoader } from "react-relay"
+import { GraphQLTaggedNode, OperationType, Variables } from "relay-runtime"
+
+interface LazyQueryRendererProps<T extends OperationType> {
+  enabled?: boolean
```
anandaroop, line 6: I don't see where `enabled` is used — does it have some implicit meaning?

src/components/LazyQueryRenderer.tsx
anandaroop: `useLazyLoadQuery (which I'm now avoiding)`

Can you elaborate on this? Is this across-the-board advice in Forque?

anandaroop (commented):
I pulled down the latest, and can no longer repro that issue 👍🏽 

Looking good, I think the PR description just needs to be updated to match reality now, re: ssr, dynamic imports, renamed component etc., along with maybe your rec about if/when to use lazy load queries.

src/components/LazyQueryRenderer.tsx
damassi: I would say so given my experience here. Reason being is that we're doing basically the same thing now as we were doing before, only slightly differently, yet its not throwing the SSR suspense error.

damassi:
Updated the readme @anandaroop and addressed the other change. Merging!

damassi:
(Or actually, i'm going to update the name of the artist autocomplete component to remove "relay" ref, and update the rest one to include "rest", since we should be using the graphql version where we can. Then i'll merge)

anandaroop:
@damassi If you promote the relay version of the component to be the default please test out the autocomplete at `/artists/dedupe` before merging 🙏🏽

damassi:
Worked like a charm @anandaroop 👍

anandaroop:
Glad to hear it! Noticed that CI is red now — that's only from the last commit yeah?

damassi:
Yep. I'm gonna quickly tackle @ovasdi's  [dev-help issue](https://artsy.slack.com/archives/CP9P4KR35/p1661958193440029) here since that's what's blowing up our tests.

damassi:
Needed to merge this, which I had lost track of: https://github.com/artsy/forque/pull/150

PR #364, opened 2022-08-31
User interests

mzikherman:
Looks like:

Screen Shot 2022-08-31 at 3 52 20 PM

The section is not included in the page at all if there are no user interests.

Opening as a draft b/c still need to add a spec.


src/pages/users/[userId].page.tsx
```diff
@@ -73,31 +73,7 @@ export const getServerSideProps: GetServerSideProps = async (ctx) => {
...
-            employer
-            profession
-            salaryUSD
-            industry
-            buyerStatus
-            priceRange
-            alternativeEmail
-            requireBidderApproval
-          }
+          ...UserSaleProfileForm_user
```
Comment (ovasdi) [inline src/pages/users/[userId].page.tsx:76]: **Question**: May I ask how you approach a refactor like this? Do you write the _interfaces/fragments_ first and then run `yarn relay` to generate the necessary _files/types_ to be referenced in your refactoring?

src/pages/users/components/form/UserSaleProfileForm.tsx
```diff
@@ -8,332 +8,355 @@ import {
...
-      {({ handleChange, handleBlur, errors, values, touched }) => {
-        return (
-          <Form>
-            <Spacer my={4} />
-            <Checkbox selected={values.requireBidderApproval}>
-              Bidder Approval Required
-            </Checkbox>
-            <Spacer my={6} />
+    <>
+      <Formik<NonNullable<UserSaleProfileForm_user$data["saleProfile"]>>
```
ovasdi, line 65: Oh, nice way to use those generated types in the form.

ovasdi (approved):
LGTM. Great refactor + learnings. 🙌

mc-jones (approved):

mzikherman:
Added a spec so this should be ready.

src/pages/users/[userId].page.tsx
Comment (mzikherman) [reply src/pages/users/[userId].page.tsx:76]: Pretty much! `yarn relay` is running in guard mode when running `yarn dev`, so you don't even need to run it separately.

Once VSCode picks up the new generated files, it's great about telling you if something isn't quite lined up as well, so if you get confused ("is it `props`, or `props.user`? whats `data` vs `props`" etc.) can get you unstuck pretty quickly.

PR #366, opened 2022-08-31
chore(misc): Update ArtistAutocomplete, add Relay test example, add utils folder 

damassi:
PR does a couple things:
- Updates ArtistAutocomplete to use more modern `matchConnection` field 
- Adds an example relay test to `pages/artist/__tests__/ArtistName.spec.ts`
- Adds a `utils` folder

Regarding the last bullet point, we used to have a `hooks` folder, but then I added a non-hook util, which meant adding another directory, `utils`. This is on top of the `test-utils` folder. So three folders all relating to "utils". Rather than that, created a `utils` folder, moving `hooks` and `test-utils` into it. This allowed us to collapse everything into one-top level folder and flatten things out.


src/pages/artist/__tests__/ArtistName.spec.tsx
```diff
@@ -0,0 +1,28 @@
+import { graphql } from "react-relay"
+import { screen } from "@testing-library/react"
+import { setupTestWrapper } from "utils/test/setupTestWrapper"
+import { ArtistName } from "../ArtistName"
+import { ArtistNameTestQuery } from "__generated__/ArtistNameTestQuery.graphql"
+
+describe("ArtistName", () => {
+  const { renderWithRelay } = setupTestWrapper<ArtistNameTestQuery>({
```
damassi, line 8: Here's how we can test relay fragments.

Still need to figure out how to test things like `useLazyLoadQuery`, `usePreloadedQuery` and `useLoadQuery`, which are all top-level graphql requests.

src/pages/artist/__tests__/ArtistName.spec.tsx
```diff
@@ -0,0 +1,28 @@
...
+    Component: (props) => <ArtistName artist={props.artist!} />,
+    query: graphql`
+      query ArtistNameTestQuery @relay_test_operation {
+        artist(id: "andy-warhol") {
+          ...ArtistName_artist
```
damassi, line 13: Note that we setup the main query, and then pass in the component fragment we want to test here

src/utils/extractNodes.ts
```diff
@@ -1,4 +1,4 @@
-export function useExtractNodes<Node extends object, T = Node>(
+export function extractNodes<Node extends object, T = Node>(
```
damassi, line 1: This is what prompted the `utils` folder creation. 

It was originally named `useExtractNodes` so that it could comfortably live in a top-level `hooks` folder without needing to add an additional `utils` folder, but then I needed to use it in a way that broke the rules of hooks. Then I realized my original mistake: it's naive to think we don't need a utils folder and that everything can just be a hook. Hooks can be rigid.

artsyjian (approved):
thanks @damassi !

PR #370, opened 2022-09-01
chore: refactor CollectorProfileForm to use a fragment + basic test

mzikherman:
This moves the fragment for the `CollectorProfile` tab to that component, and also adds a basic Relay testing spec for the form (which could probably be added to, maybe assertions on the mutation or rendering the checkbox state properly?).

Otherwise it continues to work as-is....which leads me to discover and ticket the following: https://artsyproduct.atlassian.net/browse/PLATFORM-4546 (it actually isn't displaying the right user's data!)


src/pages/users/components/form/CollectorProfileForm.tsx
```diff
@@ -110,7 +130,7 @@ export const CollectorProfileForm: React.FC<UserProps> = ({ user }) => {
                     name="collectorProfile.companyName"
                     title="Company Name"
                     placeholder="Enter Company Name"
-                    value={values.collectorProfile.companyName}
+                    value={values.collectorProfile.companyName!}
```
mzikherman, line 133: These are annoying (adding the `!`). Type-wise, seems like Formik prefers to use `undefined` rather than `null` (whereas in our case emitted Relay types from our graph are riddled with `| null`), which makes re-using the Relay types as form types tough without adding these in some places.

But, that's all to say that this is just a type thing, it's not a functional change.

src/pages/users/components/form/__tests__/CollectorProfileForm.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+        }
+      }
+    `,
+  })
+
+  it("displays the basic collector profile data", () => {
+    renderWithRelay({
+      User: () => ({
+        collectorProfile: {
+          companyName: "Catty Company",
```
mzikherman, line 23: I'm incredibly happy to say that copilot recommended this fixture 😻

ovasdi (approved):

PR #372, opened 2022-09-07
feat: add orders tab with exchange link in it

kajatiger:
There is still the question of where to put the link. 

- need to make sure exchange url is the right one for every env
- need to make sure user id is the buyer id
Bildschirmfoto 2022-09-07 um 18 58 54


.env.development
```diff
@@ -3,6 +3,7 @@
 ############################
 GRAVITY_URL="https://stagingapi.artsy.net"
 METAPHYSICS_URL="https://metaphysics-staging.artsy.net"
+EXCHANGE_URL="https://exchange-staging.artsy.net"
```
damassi, line 6: Alphabetical pls 🙏

README.md
```diff
@@ -60,6 +60,8 @@ yarn relay --watch
 
 - If you get a `Only absolute URLs are supported` upon login this may mean that you don't have a properly configured env file. Be sure to follow the setup steps above.
 
+- If you can't see certain pages (like the users page etc.), you might need to add missing user roles to your user with the assignment and the add method like so `your_user.roles = your_user.roles.add(role)` and don't forget to hit `save!` and login again.
```
damassi, line 63: Can we add the full gravity console steps here in a complete codeblock? from logging into hokusai on down.

.env.development
kajatiger: Yes! But this is currently also not working and I am trying to figure out why..

src/pages/users/[userId].page.tsx
```diff
@@ -38,6 +39,9 @@ const User: React.FC<UserProps> = ({ user }) => {
         <Tab name={"Sales Profile"}>
           <UserSaleProfileForm user={user} />
         </Tab>
+        <Tab name={"Orders"}>
```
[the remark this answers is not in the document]
Comment (kajatiger) [reply src/pages/users/[userId].page.tsx:42]: Just wondering @damassi, all the other Tab names are also wrapped in curlies... For consistency I would either unwrap them all or leave the orders one wrapped. Why did people wrap them before if that is unnecessary?

Comment (ovasdi) [inline src/pages/users/[userId].page.tsx:42]: **Question** I thought this would live under "Purchases" tab together with the Artwork Sales ([related ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-4437)). Did we decide to split them up or place these related components under "Orders"?

Comment (kajatiger) [reply src/pages/users/[userId].page.tsx:42]: Where is this purchase tab? It is not on main yet is it? Can you link to the PR/branch so I could possibly rebase my branch off of a branch that adds the purchase tab? There was no link to the PR in the ticket.

Comment (kajatiger) [reply src/pages/users/[userId].page.tsx:42]: I thought that we are going to add a list of orders pulled from exchange via metaphysics in the future and that's why I added the orders tab.

Comment (ovasdi) [reply src/pages/users/[userId].page.tsx:42]: > Where is this purchase tab? It is not on main yet is it? Can you link to the PR/branch so I could possibly rebase my branch off of a branch that adds the purchase tab? There was no link to the PR in the ticket.

The Purchases tab is not yet implemented. I was just slightly confused as both tasks are related and mention the "Purchases" tab. Just wanted to clarify.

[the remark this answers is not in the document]
Comment (damassi) [reply src/pages/users/[userId].page.tsx:42]: Best answer I can come up with is: because you can? And then its just copy pasta all the way down

src/pages/users/[userId]/components/UserNav.tsx
```diff
@@ -25,6 +27,13 @@ export const UserNav: React.FC = () => {
       <RouteTab name="Saves" href={`/users/${userId}/saves`}>
         Saves
       </RouteTab>
+      <RouteTab
+        name="Orders"
+        href={`${publicRuntimeConfig.NEXT_PUBLIC_EXCHANGE_URL}/admin/orders?q%5Bbuyer_id_eq%5D=${userId}`}
```
Comment (damassi) [inline src/pages/users/[userId]/components/UserNav.tsx:32]: @kajatiger - the issue was that you were doing 
```
publicRuntimeConfig.EXCHANGE_URL
```
vs 
```
publicRuntimeConfig.NEXT_PUBLIC_EXCHANGE_URL
```

damassi:
@kajatiger - with https://github.com/artsy/forque/pull/381 landing I went ahead and pushed to your PR (since the change required some significant rebasing). Also slightly updated it since we can now just call external links right off the tab bar, which means we don't need to create extra components.

PR #375, opened 2022-09-07
feat: add tab for user favorites

mzikherman:
Looks like:

Screen Shot 2022-09-07 at 3 39 28 PM


src/pages/users/components/UserSaves.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+interface UserSavesProps {
+  user: UserSaves_user$key
+}
+
+export const UserSaves: React.FC<UserSavesProps> = ({ user }) => {
+  const { savedArtworksConnection } = useFragment(
+    graphql`
+      fragment UserSaves_user on User {
+        savedArtworksConnection(first: 100) {
```
mzikherman, line 22: At some point (maybe in this PR?), we should do proper pagination like we have in Force, although that could also be a subtask/new task to introduce to one of these places we're doing this.

mzikherman: Oh, we do have that component in Forque already: https://github.com/artsy/forque/blob/97a4e15fd46376cd969e5294df02b9f9fb7329ec/src/components/ListPagination.tsx

Nevermind! We should be using it to start with, rather than `first: 100`. I'll update to use pagination in this PR for this component, and I'll create a ticket to update the `first: 100` locations (I can think of at least one other one) to use proper pagination.

mzikherman: I have that working locally, but needed https://github.com/artsy/metaphysics/pull/4366 - once that's merged and schema updated I'll rebase and push the pagination commit here.

Updating other locations within `User` should be straightforward.

mzikherman: Updated w/ pagination.

ovasdi (approved):
Awesome. Great to see more usage of the `useRefetchableFragment`.👏

PR #376, opened 2022-09-07
feat: add user follows tab

mc-jones:
This PR adds a `follows` tab to the `Users`, showing the user's artist and gene follows. 

### Considerations
We aren't able to get full parity with the `admin.artsy.net` follows functionality, as there is no `Artwork` app equivalent in `forque`. Should this link to somewhere else?

### Follows ups
- Add proper testing (punting a bit until we get user feedback). [Ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-4515).
- Proper pagination. [Ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-4568)
- [Maybe] Add sorting for follows in `metaphysics` and `gravity`. [Ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-4434)

### Related PR(s)
- https://github.com/artsy/metaphysics/pull/4364

Screen Shot 2022-09-07 at 4 00 33 PM

[PLATFORM-4434]

[PLATFORM-4434]: https://artsyproduct.atlassian.net/browse/PLATFORM-4434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/users/components/UserFollows.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const artistConnectionEdges = follows?.artistsConnection?.edges
+  const geneConnectionEdges = follows?.genesConnection?.edges
+
+  const alphabatizedArtists = orderBy(
+    artistConnectionEdges,
+    ["node.name"],
+    ["asc"]
+  ).map((el, idx) => {
+    const currentNode = el?.node
+    return <Text key={idx}>{currentNode?.name}</Text>
```
mzikherman, line 45: the `key` should maybe be the `currentNode?.id` or something? (not sure, #minor cc @damassi )

src/pages/users/components/UserFollows.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }
+      }
+    `,
+    user
+  )
+
+  const artistConnectionEdges = follows?.artistsConnection?.edges
+  const geneConnectionEdges = follows?.genesConnection?.edges
+
+  const alphabatizedArtists = orderBy(
```
mzikherman, line 39: Rather than building up JSX and rendering via `{alphabatizedArtists}` (typo btw!), it's more canonical to have a component.

Something like:

``` tsx
const AlphabetizedGenesList = () => {
    return (
      <>
        {orderBy(geneConnectionEdges, ["node.name", ["asc"]]).map((el, idx) => {
          const currentNode = el?.node
          return <Text key={idx}>{currentNode?.name}</Text>
        })}
      </>
    )
  }
```

then, in the markup for the layout / where this list is being used:

``` tsx
        <Column span={6}>
          <AlphabetizedGenesList />
        </Column>
```

mzikherman (approved):
LGTM! It's a super bare-bones display so I have to imagine we _could_ do better if needed: sub-tabs under 'Follows' for 'Artists, 'Genes', 'Profiles', each of those rendering that type in a paginated list view, etc. - but if admins find this kind of thing useful as they use Forque/Torque for user management + lookups, cool! 👍 

I would add a simple spec before merging even if it's an MVP. Coming back to add tests is something I tell myself on the order of "I'm only going to eat one cookie per day" when I'm at the grocery store and convincing myself to buy a thing of salted caramel choc. chip!

src/pages/users/components/UserFollows.tsx
```diff
@@ -0,0 +1,86 @@
...
+              }
+            }
+          }
+        }
+      }
+    `,
+    user
+  )
+
+  const artistConnectionEdges = follows?.artistsConnection?.edges
```
damassi, line 36: We should be using the `extractNodes` function whenever we need to get nodes and edges

src/pages/users/components/UserFollows.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const geneConnectionEdges = follows?.genesConnection?.edges
```
damassi, line 37: ditto here

src/pages/users/components/UserFollows.tsx
damassi: ✔️

damassi: (Or even just `GenesList` -- internal implementation details don't necessarily need to be encoded in the name as they might change)

src/pages/users/components/UserFollows.tsx
damassi: I pretty much always just use the index (as its done here) and have never really run into trouble but someone as meticulous as Damon would key it on `id`. Who do you wanna be @mc-jones 😆

src/pages/users/components/UserFollows.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <Column span={6}>{alphabatizedGenes}</Column>
+        ) : (
+          <Column span={6}>
+            <Text>User does not follow any Genes</Text>
+          </Column>
+        )}
+      </GridColumns>
+    </>
+  )
+}
```
damassi, line 86: Second @mzikherman here - better to use real components rather than interpolated variables

src/pages/users/components/AlphabetizedFollowsList.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+interface followsConnection {
+  followsList: {
+    readonly id: string
+    readonly name: string | null
+  }[]
+  followsType: "Artists" | "Genes"
+}
+
+export const AlphabetizedFollowsList: React.FC<followsConnection> = ({
```
damassi, line 13: No need for the word "Alphabetized" to be in the component name; it's an implementation detail.

mc-jones:
>LGTM! It's a super bare-bones display so I have to imagine we could do better if needed: sub-tabs under 'Follows' for 'Artists, 'Genes', 'Profiles', each of those rendering that type in a paginated list view, etc. - but if admins find this kind of thing useful as they use Forque/Torque for user management + lookups, cool! 👍

I agree with you here. I anticipate there are going to be changes needed once we get user feedback here, so I erred on the side of keeping it super simple for now, rather than needing to revert back to unnecessary UX at this stage. I wouldn't feel comfortable releasing this into production, but think it is a good first baby step and sets up a basis for the next iteration.

> I would add a simple spec before merging even if it's an MVP. Coming back to add tests is something I tell myself on the order of "I'm only going to eat one cookie per day" when I'm at the grocery store and convincing myself to buy a thing of salted caramel choc. chip!

😆 Take this point and have added a simple spec. Just for full transparency, I definitely recognize this is not an ideal approach and have likely punted on test beyond what is reasonable for the size of work, even if an MVP, especially with the relay testing utils added in by @damassi.

src/pages/users/components/__tests__/UserFollows.spec.tsx
```diff
@@ -0,0 +1,50 @@
...
+          artistsConnection: {
+            edges: [
+              {
+                node: {
+                  id: "123",
+                  name: "Catty Artist",
+                },
+              },
+            ],
+          },
```
mc-jones, line 32: I'm seeing some weird behavior when I expand this test to include two nodes on the `artistsConnection`, which is causing inconsistent test results (it always works with just one node in each `connection`). In some cases it was rendering the second `artistsConnection` node twice and the first node zero times, In others it was merging the `genesConnection` node into `artistConnection`. 

I need to look into this more.

damassi: It might be worth just mocking out the `node` type - like `Artist: () => ({ name: ... })` and seeing if its populated automatically from the test utils as they mock the edges

mc-jones: ...oh dear. I didn't realize I failed to change the id on the node when adding the second one. This is my sign I need to clock out... Nothing to see here.

PR #380, opened 2022-09-08
feat(users): Add UserSearch component

damassi:
This adds a new AutocompleteInput to the /users page, as a quick way to get to a user when you don't know their internalID. Takes advantage of our new `<LazyQueryRenderer />` component, too. 

email-input

Also updates `LazyQueryRenderer` with a couple more props:
- `enabled: boolean`: only perform fetch if certain conditions are met 
- `fetchPolicy`: allows you to update the relay fetch policy for the query


src/pages/users/components/UserSearch.tsx
```diff
@@ -0,0 +1,54 @@
...
+        query UserSearchQuery($email: String!) {
+          user(email: $email) {
+            name
+            internalID
+          }
+        }
+      `}
+      variables={{ email }}
+      enabled={/\S+@\S+\.\S+/.test(email)} // only fetch if email
+      render={({ isLoading, data }) => {
```
ovasdi, line 24: **Question**: so this `render` is a callback that executes when `LeazyQueryRenderer` returns a fragment with two props, `isLoading` and `data`, those are then passed down to `AutocompleteInput` which is what ends up getting rendered by this component. Is that correct?

ovasdi (approved):
Great stuff. ✨

mc-jones:
Thanks for this, @damassi! 💯 🙏

src/pages/users/components/UserSearch.tsx
damassi: Yup, if you're rendering `LazyQueryRenderer`, `render` always gets called and attempts to render the component whether `isLoading` is true or false. What this does in addition tho is to report `isLoading` status -- which you can't typically get in a code-usable way via react suspense.

PR #386, opened 2022-09-09
chore: fix couple Head titles

ovasdi:
While reviewing the changes in https://github.com/artsy/forque/pull/381 noticed a couple of tiny inconsistencies regarding component head titles.


damassi:
👍

PR #389, opened 2022-09-12
refactor: users folder structure

mc-jones:
This PR refactors the users app folder structure a bit. Specifically it:

1. Deletes the `form` suffix from components. While many of the components use a form, they are often mixed with read-only data. The `form` suffix didn't seem to add anything here and seems to be a holdover from the initial work on which this was based. 
2. Deletes the `form` subdirectory and moves the components one level up into the `...users/[userId]/components` directory, for the reasons mentioned above in point 1. Additionally, this flattens out the folder structure, and I would argue avoids unnecessary nesting. 
3. Deletes unused components from the initial work (see #184)

ToDo: 

- [x] rename fragments


damassi (approved):
👌

PR #393, opened 2022-09-13
fix(tests): add more inclusive jest regex

damassi:
We noticed that some tests weren't matching but this seems to fix it!


damassi:
_no go_

PR #399, opened 2022-09-14
feat: add pagination to user interests and follows

mzikherman:
I had to split up the `UserFollows` component (which fetched artists and genes together) further, since each one can be paginated individually w/ their own params (including sorts if we add that, etc.).

Otherwise, pretty vanilla drop-in!


src/pages/users/[userId]/index.page.tsx
```diff
@@ -36,9 +36,6 @@ export const getServerSideProps: GetServerSideProps = async (ctx) => {
           ...UserLayout_user
 
           ...UserAccount_user
-          ...UserCollectorProfile_user
```
Comment (mzikherman) [inline src/pages/users/[userId]/index.page.tsx:39]: oh, incidentally I noticed this was still being fetched in this fragment, but these are split into their own page and `getServerSideProps` invocation, so removed them.

mc-jones (approved):
🙏

PR #409, opened 2022-09-19
chore(storybook): Remove storybook

damassi:
Since we've already got palette (and palette's storybook) and Forque is basically a light composition of palette components, removing it from forque to remove the number of dependencies we need to worry about. 

Opening this PR and assigning it to you @mc-jones to make the call!


mc-jones:
All for reducing the number of dependencies if this isn't providing significant value. 

👍 🙏

PR #411, opened 2022-09-20
feat: user purchased artworks open in new tab

ovasdi:
This opens user-purchased artworks in a new tab (when clicking on purchased artwork). It seems like a better UX, but if others think it isn't, I can leave it as is.


damassi (approved):
Def better UX!

